PR #43, opened 2021-11-11
README.md - ensure all work has a link

david-a-wheeler:
We have some best practices work not linked, or the links aren't good ones.
Let's fix that.

Also: Everything is continuing in some sense, so I instead added "(incubating)" to show drafts not ready to use.


david-a-wheeler:
@SecurityCRob - Hopefully this is good to merge!

PR #53, opened 2022-03-03
Add OWASP MASVS and MSTG for mobile application security

cpholguera:
The OWASP MASVS and MSTG are the place to go for developing and testing secure mobile applications. The MASVS is OS-agnostic and the MSTG goes into detail on each platform (Android and iOS).


cpholguera:
Thank you @SecurityCRob!

PR #76, opened 2022-08-28
Use curly quotes instead of straight quotes

david-a-wheeler:
Use UTF-8 curly quotes instead of straight quotes
in the two concise guides.


ran-dall (approved):
LGTM

PR #77, opened 2022-08-28
Fix crosslinking URLs in "Concise" guides

david-a-wheeler:


david-a-wheeler:
This is a trivial hyperlink fix, and we don't want people to go to the old Google docs files, so I intend to merge this quickly.

PR #86, opened 2022-09-01
Note that tools/services listed are merely examples

david-a-wheeler:


david-a-wheeler:
@SecurityCRob - I hope you don't mind, but I created a proposed solution for the text headers, noting that tools/servers are merely examples. With this change, I think all WG issues are resolved.

PR #98, opened 2022-10-15
Update to OpenChain Security Assurance specification 1.1

david-a-wheeler:
Update the OpenChain Security Assurance specification to version 1.1, and tweak the language to make it a little clearer what its role is. The specification focuses on how organizations should be looking for *known* vulnerabilities, which wasn't as clear in the concise guide, so I hope this makes it better.


david-a-wheeler:
I'd like Shane Coughlan's review @shanecoughlan before merging this. This document should reference the OpenChain document, but put it in a proper context so it's not confusing to readers.

david-a-wheeler:
[shanecoughlan](https://github.com/shanecoughlan) - thanks for the review!

I think we should merge this soon. The concise guide for evaluation of OSS currently references the old name and an old (and no longer working URL); this PR updates them to the current document name and URL. So it's not really changing anything significant about the OpenSSF guide, merely updating the OpenSSF guide.

Any other comments?

shanecoughlan:
Good to go on my side. :)

david-a-wheeler:
I'm going to merge this, since otherwise we point to the wrong (old) name and the wrong (old) URL, which helps no one.

PR #100, opened 2022-11-07
Add the GitHub Super Linter for Markdown as a GitHub Action

### LINKED ISSUE #74: Enforce a consistent markdown style
Currently, we don't enforce any consistent markdown style. 

I think it would make sense to add [Markdownlint](https://github.com/DavidAnson/markdownlint) via [GitHub's Super-Linter](https://github.com/github/super-linter) in a GitHub Action.

All current documents can be auto-reformatted to comply once an agreed Markdownlint configuration is established.

ran-dall:
This adds the GitHub Super Linter for Markdown as a GitHub Action.

It should be noted that the GitHub Action is set to only check new files in a commit/PR.

Closes #74

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,0 +1,2 @@
+# Default state for all rules
+default: true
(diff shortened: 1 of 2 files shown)


ran-dall:
I made it a draft because I'm still running a few tests, and intend to also make another PR with the markdown fixes.

PR #113, opened 2023-02-17
Add initial draft of Toolchain Hardening for C and C++ guide

thomasnyman:
As discussed during the C/C++ Compiler Options BP Guide call on 2023-02-15 I'm opening this draft PR in order to share the initial Markdown draft of the "Toolchain Hardening for C and C++" guide from Ericsson. There were already good suggestions for improvement during the call. We can use this PR to continue the discussion and start incorporating changes.

### DIFF
--- a/docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -1,10 +1,10 @@
-## Toolchain hardening for C and C++
+# Compiler Options Hardening Guide for C and C++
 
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
I have a number of comments, but I'm also really excited to see this. THANK YOU!

I'm very concerned about the term "Toolchain Hardening". We have a *lot* work going on to counter supply chain attacks, in part due to serious problems like the attack on SolarWinds' Orion. It's true that some people use that phrase, but *also* is used to describe hardening of build environments against external attacks, so it's a really ambiguous term.

I think the title should at least include the term "option" or "options", and maybe "option flags". These are option flags, if we put them in the title everyone will understand. Also, toolchains are *lot* more than compilers. Make and cmake are part of many toolchains, for example. So I'd use the term "compiler" - that's what everyone calls them.

So I'd title this something like this:

> Compiler option hardening guide for C and C++

Feel free to move around the order of words, etc. But I think if "compiler", "option", "C" and "C++" are in the title, people are more likely to find it *and* will understand what they found :-).

david-a-wheeler:
FYI: We prefer markdown to meet certain lint requirements - you can see in "details" all the things it whines about :-).

We can forcibly merge it & then fix that stuff, if desired. I actually think that would be better - we do want to fix the problems, but the sooner we make this visible to others as a starting point, the better.

PR #116, opened 2023-02-28
Remove dates from Concise Guides

david-a-wheeler:
The concise guides are already under version control, we don't need a date (that will itself go out of date).


david-a-wheeler:
This was agreed on today in the Best Practices WG meeting. Implemented in the same day!

PR #117, opened 2023-02-28
Clarify licenses (software Apache 2.0, docs CC-BY-4.0)

david-a-wheeler:
The OpenSSF charter says this, but it's not as obvious here. Make it obvious.


david-a-wheeler:
@SecurityCRob - I believe this simply documents what's in the charter & agreed on by the WG. We discussed this today at the OpenSSF Best Practices WG, and there seemed to be consensus. So I'm going to just merge this. If that's an error, please let me know & we can revert this.

david-a-wheeler:
I wanted to make sure this was clear *before* tomorrow's meeting about the C/C++ Compiler Options. We told them the result would be CC-BY-4.0, and that *is* what the charter says, but somehow that didn't get into this repo.

SecurityCRob:
agreed

Cheers,

CRob
Director of Security Communications
Intel Product Assurance and Security

PR #118, opened 2023-03-01
Add sample hyperlink for -fstack-protector-strong

david-a-wheeler:
It would be nice if readers could click on a compiler flag in a summary and immediately see the details about that flag.

I propose adding hyperlinks to implement that. It would make the markdown a *little* messier, but not outrageously so, and it would make the document *much* easier to navigate. I think we need to make it *easy* for readers to get from a summary to the details they need, or they won't read the details they need.

This adds an internal hyperlink from a flag description in a summary to an anchor on the details. The anchor is a span with id; I could have used a name="..." instead, but many seem to be preferring id so I used that. The id is exactly the flag name (e.g., `-fstack-protector-strong`); leading dashes *are* allowed in ids.

NB: The sample entry makes 1 table entry longer than the others; if they *all* do that, then columns could be re-aligned again without lots of space.


david-a-wheeler:
Note: Markdown errors will continue until we fix the document. @ran-dall is working on that.

gkunz:
I think the benefits outweight the downsides of having more ugly markdown.

It would be great to have also link-checking as part of the linting. Can the GitHub linter do that?

david-a-wheeler:
@gkunz - yes, our markdownlint checker can internal links. I can attest that in part because I've had to fix them :-). We should probably double-check its configuration to ensure it actually *checks* the links :-). That'll be easier to verify once we fix the initial problems.

PR #123, opened 2023-03-07
Fix markdown in compiler hardening guide

david-a-wheeler:


david-a-wheeler:
@ran-dall - I hope you don't mind, but I took a stab at fixing the markdown in the compiler guide draft so that it passes our linter.

Everyone: Comments?

david-a-wheeler:
This one has extra spaces after `=`, but so does the current version in `main`. This seems to be a weakness of GitHub markdown. That may be fixable, but I'd like to get rid of the markdown warnings ASAP, then see if we can improve the display.

ran-dall (approved):
LGTM

gkunz (approved):

PR #128, opened 2023-03-15
Add __GLIBCXX_ASSERTIONS

david-a-wheeler:
I didn't find a good source for when these were added to gcc/clang, so I left that as "?".


docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -89,6 +89,7 @@ Table 2: Recommended compiler options that enable run-time protection mechanisms
 | Compiler Flag                                                                             |            Supported by            | Description                                                                                  |
 |:----------------------------------------------------------------------------------------- |:----------------------------------:|:-------------------------------------------------------------------------------------------- |
 | [`-D_FORTIFY_SOURCE=2`](#-D_FORTIFY_SOURCE=2) <br/>(requires `-O1` or higher)             |      GCC 4.0<br/>Clang 5.0.0       | Fortify sources with compile- and run-time checks for unsafe libc usage and buffer overflows |
+| [`-D__GLIBCXX_ASSERTIONS`](#-_GLIBCXX_ASSERTIONS)                                         |      GCC ?<br/>Clang ?             | (C++ only) Run-time bounds checking for C++ strings and containers; can impact performance.  |
```
siddhesh, line 92: The macro has a single preceding underscore: `_GLIBCXX_ASSERTIONS`.

siddhesh:
> I didn't find a good source for when these were added to gcc/clang, so I left that as "?".

It was introduced first in gcc 6: https://gcc.gnu.org/gcc-6/changes.html

docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Thank you!! It was right in some places but not others. Fixed.

david-a-wheeler:
Thanks, I've added GCC 6.0. I still don't know when clang added it - anyone know a good source for this?

ran-dall:
@david-a-wheeler Did we want also to point out here that if it doesn't compile with `_GLIBCXX_ASSERTIONS`, it can be considered a bug?

david-a-wheeler:
> @david-a-wheeler Did we want also to point out here that if it doesn't compile with _GLIBCXX_ASSERTIONS, it can be considered a bug?

No - or at least, I wouldn't make it specific to this flag.

*I* think code should compile & run with all of these enabled. However, developers may decide that a specific flag (no matter which one it is) is okay for their circumstance. So the issue is not specific to this flag.

ran-dall:
> I think code should compile & run with all of these enabled. However, developers may decide that a specific flag (no matter which one it is) is okay for their circumstance. So the issue is not specific to this flag.

So is our unified position that it can be considered a bug if something doesn't compile with any of our recommended flags enabled?

thomasnyman:
> 



> Thanks, I've added GCC 6.0. I still don't know when clang added it - anyone know a good source for this?

Since this is a feature of GCC’s implementation of the C++ standard library (libstdc++) it is the version of GCC available that matters even if the feature is used through Clang as long as Clang has been [instructed to use libstdc++](https://releases.llvm.org/16.0.0/tools/clang/docs/Toolchain.html#id6).

LLVM also provides its own version of the C++ standard library ([libc++](https://libcxx.llvm.org/)) which provides an equivalent `_LIBCPP_ASSERT` macro. Marshall Clow talks about it here: https://www.youtube.com/watch?v=1iHs_K2HpGo&t=990s 

This was recently [audited to only use constant-time conditions in the libc++ code base](https://github.com/llvm/llvm-project/commit/c87c8917e3662532f0aa75a91caea857c093f8f4) so as far as I can tell it is equivalent to the libstdc++ macro.

`_LIBCPP_ASSERT` was first introduced in https://github.com/llvm/llvm-project/commit/92bd7c0337b4e63ea823573f3ce92eb6afbb082c meaning it would have first appeared in [LLVM release 3.3.0](https://github.com/llvm/llvm-project/releases/tag/llvmorg-3.3.0).

thomasnyman:
Merged as per discussion during C/C++ Compiler BP Guide on [2023-04-12](https://docs.google.com/document/d/1x-fMIt6ZIV6SM-M29xEVKeB505GbDzSIVXg-8uF_FQU/edit#bookmark=id.q82jhtsux5i6).

PR #131, opened 2023-03-15
Tweak introduction of compiler options guide

david-a-wheeler:
This adds some information from the old version, and makes it clear we encourage developers to treat not supporting these compiler options as a problem.


ran-dall (approved):
LGTM

gkunz (approved):

PR #132, opened 2023-03-15
Add `gcc -v` section in Complier Hardening Guide

### LINKED ISSUE #119: Distros' Default C/C++  Complier Options
As discussed on today's (3/1/23) phone call, I'm putting together a paper with the default compiler options used by each independent Linux distro.

https://docs.google.com/document/d/1QGyDVgu0bGdKkdSIIJMHW35R044vGLa6-9tVs86A4_8/edit?usp=sharing

ran-dall:
This adds a section about using `gcc- v` to examine certain GCC options.

I also say this closes #119


gkunz:
Out of curiosity, I ran this command on an Ubuntu 20.04.06 LTS system with the following output:

`
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/9/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:hsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 9.4.0-1ubuntu1~20.04.1' --with-bugurl=file:///usr/share/doc/gcc-9/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,gm2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-9 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-9-Av3uEd/gcc-9-9.4.0/debian/tmp-nvptx/usr,hsa --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1)
`

It is not straightforward to me to determine from this output which options are in fact enabled or disabled. I see "--disable-werror" and "--enable-default-pie" for example, but it is not clear to me (and maybe other readers) which of the recommended options are covered.

Should we add more details around this (or does Ubuntu not enable many options by default)?

ran-dall:
> It is not straightforward to me to determine from this output which options are in fact enabled or disabled. I see "--disable-werror" and "--enable-default-pie" for example, but it is not clear to me (and maybe other readers) which of the recommended options are covered.
>
> Should we add more details around this (or does Ubuntu not enable many options by default)?

@gkunz I see where you're coming from. For example, building GCC with  `--enable-default-pie` will enable the `-fPIE` build flag and `-pie` linker flag, which are on the guide. 

Perhaps adding a column (next to Complier Flag) with these options that enable stuff is applicable and would be helpful for the reader to make the association?

gkunz:
Hi @ran-dall, thinking about this a bit, I tend to say that we should just keep the text you add in this PR and skip the additional column with build / configuration options. I think this opens yet another dimension which is likely not very relevant to the primary target audience (developers + distributors), i.e., I don't think that many of them first build their compiler and then their app (except for Gentoo folks, of course :-) ).

So, in a nutshell, include your pointer to the build-time options, but don't go into details for the sake of clarity of the guide.

gkunz (approved):

ran-dall:
> Hi @ran-dall, thinking about this a bit, I tend to say that we should just keep the text you add in this PR and skip the additional column with build / configuration options. I think this opens yet another dimension which is likely not very relevant to the primary target audience (developers + distributors), i.e., I don't think that many of them first build their compiler and then their app (except for Gentoo folks, of course :-) ).

@gkunz I agree; however, as I said on the call, from what I see in the greater OSS community, folks don't use the `gcc` community container but rather elect to pull down images such as `debian` or `alpine` (in their CI/CD or build pipelines) and install `gcc` with the local package manager which will deploy an instance of `gcc` configured by that respective packaging team.

Another suggestion could be to clarify this a bit more in the originally added language that if you get `gcc` from pretty much anywhere other than `gcc`, it might already have certain flags enabled/disabled.

david-a-wheeler:
> Another suggestion could be to clarify this a bit more in the originally added language that if you get gcc from pretty much anywhere other than gcc, it might already have certain flags enabled/disabled.

That seems like the better approach.

Also, "default flags" is more complicated. For example, if in Fedora you run gcc, those aren't the same as the "default flags used by Fedora packages" because the packaging system has its own defaults (set by the distribution).

I don't know how common Linux or BSD distributions are in the embedded world. I think a lot of developers (especially the under $0.10 CPU crowd) will typically use IDEs provided by the manufacturer, which may be cobbled together until they "have something that works". Without more data I would assume nothing about "what's common".

ran-dall:
> That seems like the better approach.

@david-a-wheeler @gkunz Done. If you don't mind, please review the new language, added in 54eb30b9b5eb7534fcd742a310dee117598fea56.
 
> Also, "default flags" is more complicated. For example, if in Fedora you run gcc, those aren't the same as the "default flags used by Fedora packages" because the packaging system has its own defaults (set by the distribution).

Agreed. (We haven't even touched on "default profile" which FMU is separate from these "default flags".)

> I don't know how common Linux or BSD distributions are in the embedded world. I think a lot of developers (especially the under $0.10 CPU crowd) will typically use IDEs provided by the manufacturer, which may be cobbled together until they "have something that works". Without more data I would assume nothing about "what's common".

I don't have numbers, but I know there are now projects that support many of these products with a Unix-like OS (I don't know if I'd go as far as calling it Linux or BSD). I don't know how popular they are, but I assume they're popular enough to justify their existence. Beyond that, I don't have much experience with those types of embedded systems (under $0.10 CPU).

david-a-wheeler:
I like this revision better, thanks.

david-a-wheeler:
+1

thomasnyman (approved):

PR #133, opened 2023-03-17
Fix some incorrect assumptions with _FORTIFY_SOURCE

siddhesh:
_FORTIFY_SOURCE is a compile-time + run-time feature at all levels, not just at level 1.  Add some more specific information on the differences between levels 1 and 2 and also clarify the following issues:

- The fortify abort hasn't printed a backtrace for a while now, we removed it long ago.
- There's no performance difference between fortifications level 1 and 2.
- There are other headers too that add fortification, not just stdio.h and string.h.  More importantly, we'll probably end up adding more fortification in glibc the future.


david-a-wheeler:
If FORTIFY_SOURCE has run-time implications at level 1, then it should probably be removed from table 1. Instead, have table 2 just list FORTIFY_SOURCE=3, with details about levels 1, 2, and 3 below in the the details.

siddhesh:
> OK, I'll broaden this PR to include `_FORTIFY_SOURCE=3` stuff then.

I've done this now.  Hopefully this addresses everything needed on the `_FORTIFY_SOURCE` front.

docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -239,9 +240,17 @@ Both `_FORTIFY_SOURCE=1` and `_FORTIFY_SOURCE=2` are expected to have a negligib
...
 
-However, when enabling `_FORTIFY_SOURCE=2` in existing code bases regression testing should be used to ensure the run-time checks do not adversely affect existing features.
+- If the fortified glibc function calls show up as hotspots in your application performance profile, there is a chance that `_FORTIFY_SOURCE` may have a negative performance impact. This is not a common or widespread slowdown[^29] but worth keeping in mind if slowdowns are observed due to this option
+- Applications that use the GNU extension for flexible array members in structs[^30] may confuse the compiler into thinking that an object is smaller than it actually is, resulting in spurious aborts. The safe resolution for this is to port these uses to C99 flexible arrays but if that is not possible (e.g. due to the need to support a compiler that does not support C99 flexible arrays), one may need to downgrade or disable `_FORTIFY_SOURCE` protections.
```
thomasnyman, line 246: Is this caveat specific to `_FORTIFY_SOURCE=3`? 

If so, should that be stated explicitly and the statement about downgrading say "downgrade to ´_FORTIFY_SOURCE=2´"?.

docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -213,22 +212,24 @@ For example, developers can decide to promote warnings that indicate interferenc
...
+- `-D_FORTIFY_SOURCE=1`: conservative, compile-time and runtime checks; will not change (defined) behavior of programs. Checking for overflows is enabled when the compiler is able to estimate a compile time constant size for the protected object.
 - `-D_FORTIFY_SOURCE=2`: stricter checks that also detect behavior that may be unsafe even though it conforms to the C standard; may affect program behavior by disallowing certain programming constructs. An example of such checks is restricting of the `%n` format specifier to read-only format strings.
+- `-D_FORTIFY_SOURCE=3`: Same checks as those covered by `-D_FORTIFY_SOURCE=3` except that checking is enabled even when the compiler is able to estimate the size of the protected object as an expression, not just a compile time constant.
 
 To benefit from `_FORTIFY_SOURCE` checks following requirements must be met:  
 
-- the application must be built with -O1 optimizations or higher
-- the sizes of the destination buffers must be known at compile time, and
+- the application must be built with `-O1` optimizations or higher; at least `-O2` is recommended.
```
thomasnyman, line 231: Is there a benefit to `_FORTIFY_SOURCE` behavior between `-O1` and `-O2´ or is the recommendation simply motivated by better overall performance?

siddhesh: `_FORTIFY_SOURCE` depends on object size estimation and that improves with optimization, so `_FORTIFY_SOURCE` coverage improves with better optimization.

docs/Compiler_Hardening_Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
siddhesh: It's not specific to `_FORTIFY_SOURCE=3`; it's just that applications are more likely to hit these issues with `_FORTIFY_SOURCE=3` because the coverage is significantly higher.  A number of applications disabled fortification before `3` was introduced and so far that list has not increased in my F38 testing.

thomasnyman (approved):

PR #139, opened 2023-03-29
Add -fno-delete-null-pointer-checks

david-a-wheeler:


david-a-wheeler:
I didn't have the dates when GCC and clang added this, so I left that information blank. If anyone has that information that'd be great.

Another flag we should probably add is `-fno-strict-overflow`; see Wang, Xi, et al., Undefined Behavior: What Happened to My Code?, APSys ‘12, 2012, ACM, <https://pdos.csail.mit.edu/papers/ub:apsys12.pdf>. But I think that's better handled as a separate pull request, and it would be similar to this one, so let's work on this first & once it's resolved do something similar.

david-a-wheeler:
@ran-dall @gkunz - thoughts about adding this?

06kellyjac:
> I didn't have the dates when GCC and clang added this, so I left that information blank. If anyone has that information that'd be great.


I did a bit of digging:

It seems to have lived in gcc since https://github.com/gcc-mirror/gcc/commit/b6d24183dfb4a8194e3b0357ce44723735ad0da1 ([code](https://github.com/gcc-mirror/gcc/commit/b6d24183dfb4a8194e3b0357ce44723735ad0da1#diff-9e31bd2d2496420aa34b8fc99ff6d1bf30f606032fe68eabbca9eac3c3189e69R571-R574)) under the name `delete-null-pointer-checks`

Before that flag existed it was (briefly) turned on by `optimize`
The deletion itself was added here https://github.com/gcc-mirror/gcc/commit/dfdb644f4ff71f04c78e307035d81c3fac0a18b6

I think it's safe to say it was in gcc 3.0

```
λ cd ./gcc-3.0
λ rg "delete_null_pointer_checks"
gcc/ChangeLog
3255:	* gcse.c (delete_null_pointer_checks_1): Add delete_list argument,
3257:	(delete_null_pointer_checks): Delete insns from delete_list after
3258:	all delete_null_pointer_checks_1 passes are done.

gcc/ChangeLog.3
2461:	* gcse.c (delete_null_pointer_checks): Likewise.
4459:	* gcse.c (delete_null_pointer_checks): Change cast to unsigned int*.
7067:	(delete_null_pointer_checks): Mark arg F as unused.
7692:	(delete_null_pointer_checks): Likewise.
9067:	(delete_null_pointer_checks_1): BLOCK_REG now pointer to unsigned.
14451:	(delete_null_pointer_checks): Likewise.
16056:	* gcse.c (delete_null_pointer_checks_1): Cope when

gcc/ChangeLog.1
7568:	get_condition() which is used in delete_null_pointer_checks().

gcc/ChangeLog.4
12168:	(delete_null_pointer_checks_1): Likewise.

gcc/ChangeLog.2
2595:	* gcse.c (delete_null_pointer_checks_1): Kill unused s_preds
2597:	(delete_null_pointer_checks_1): No longer need to compute the
2732:	(delete_null_pointer_checks_1): Use compute_available.
3212:	(delete_null_pointer_checks_1): New function.
3213:	(delete_null_pointer_checks): Use it.
3978:	* gcse.c (delete_null_pointer_checks): Fix typo in previous change.
4028:	* gcse.c (delete_null_pointer_checks): Only record non-null info
4086:	(delete_null_pointer_checks): Likewise.
5558:	* gcse.c (delete_null_pointer_checks): Returns void.
5688:	(delete_null_pointer_checks): Likewise.
6633:	* toplev.c (flag_delete_null_pointer_checks): New.
6636:	elimination on flag_delete_null_pointer_checks.
7038:	(delete_null_pointer_checks): Likewise.
7039:	* rtl.h (delete_null_pointer_checks): Declare.
7040:	* toplev.c (rest_of_compilation): Call delete_null_pointer_checks.

gcc/gcse.c
631:static void delete_null_pointer_checks_1 PARAMS ((varray_type *, unsigned int *,
4984:delete_null_pointer_checks_1 (delete_list, block_reg, nonnull_avin,
5155:delete_null_pointer_checks (f)
5234:      delete_null_pointer_checks_1 (&delete_list, block_reg, nonnull_avin,

gcc/rtl.h
1873:extern void delete_null_pointer_checks	PARAMS ((rtx));

gcc/toplev.c
660:static int flag_delete_null_pointer_checks;
1043:  {"delete-null-pointer-checks", &flag_delete_null_pointer_checks, 1,
2996:      /* ??? Run if-conversion before delete_null_pointer_checks,
3005:      if (flag_delete_null_pointer_checks)
3006:	delete_null_pointer_checks (insns);
3059:      if (flag_delete_null_pointer_checks)
3066:	  delete_null_pointer_checks (insns);
4774:      flag_delete_null_pointer_checks = 1;
```

I found a reference to it in gcc 2.95.1 but only in the changelog

```
λ cd gcc-2.95.1
λ rg "delete_null_pointer_checks"
gcc/ChangeLog
8726:	get_condition() which is used in delete_null_pointer_checks().
```

Based on the implementation commit being marked 1999-09-20 and gcc 2.95.1 being released 1999-08-19 I'm guessing there was some patch that made it in earlier than the other

```
λ git branch
* (HEAD detached at 3cdbd1f8ac)
  master
λ rg "delete_null_pointer_checks"
gcc/ChangeLog
4:	get_condition() which is used in delete_null_pointer_checks().
```

https://github.com/gcc-mirror/gcc/commit/3cdbd1f8ac98649ab5eaba2d9d065653e8a4820d

david-a-wheeler:
[06kellyjac](https://github.com/06kellyjac) thanks, I added "GCC 3.0" per your research. I don't have clang information, but I think we can add this as-is and add that added information later.

06kellyjac:
For clang it looks like it was added in 7.0.0

https://github.com/llvm/llvm-project/commit/da08f6ac16b6a53ab25ecb56e1be800c5f0095d9

This commit also has a nice link to a Linus Torvalds email we might want to reference https://lkml.org/lkml/2018/4/4/601

> ... we had buggy code that accessed a pointer
before the NULL pointer check, but the bug was "benign" as long as the
compiler didn't actually remove the check.
> ...
> Removing the NULL pointer check turned a benign bug into a trivially
exploitable one by just mapping user space data at NULL ...

https://github.com/llvm/llvm-project/blob/llvmorg-7.0.0/clang/include/clang/Driver/Options.td#L1086-L1088

before that it has `delete-null-pointer-checks` which you _might_ be able to force off but that digging will take a lot more effort

https://github.com/llvm/llvm-project/blob/llvmorg-6.0.1/clang/include/clang/Driver/Options.td#L2650-L2651

david-a-wheeler:
Lots of discussions today about this. It might be useful to create a third table for "options that reduce the impact of a vulnerability, but also hide problems, it'd be better to fix the problem" or something like that. Sadly, in many cases there are no options to detect the problem, & it's better to reduce the impact of a problem than just let the problem fester.

david-a-wheeler:
I think we *should* add `-fno-delete-null-pointer-checks`.

This option is currently used when compiling the Linux kernel. This was added after this example: <https://isc.sans.edu/diary/A+new+fascinating+Linux+kernel+vulnerability/6820>. There was a vulnerability that led to total bypass, including a bypass of SELinux, because null pointer checks were silently removed. This is an unfortunately easy and common problem.

The argument against this option and similar options is that "it reduces the impact of a vulnerability, but also hides problems". When you're running production code, hiding problems is not a *problem*, that is a *benefit*. I think those protesting are missing that the current document is focused on creating *production* code. I *agree* that you want to run tests *without* this and similar flags, to maximize bug-catching. But production use  is not the appropriate time to increase the likelihood of serious vulnerabilities.

In a later version of this document we should expand the scope to providing option flags for builds that maximize error detection during tests. In *those* cases we should append additional flags that turn on sanitizers and disable some last-minute mechanisms, e.g., use `-delete-null-pointer-checks`. In short, testing should involve tests of two different kinds of builds: (1) builds for maximizing detection of problems and (2) the "real" build that you intend to use in production. But let's make sure we reduce the likelihood of vulnerabilities in the production system.

david-a-wheeler:
I think we've removed a blocker. We've modified the document to clearly distinguish between *production code* and *instrumented test code*.  The comments earlier that "these options would quietly hide defects" are true, but you *want* to remove defects from production code. The (legitimate) concern is that you want to *reveal* these problems when running instrumented test code. So let's add a new table entry "production code" and add these issues. I'll revise the text to match.

david-a-wheeler:
This has become messy due to merge conflicts. I'll create another PR so we have a clean change.

david-a-wheeler:
I've rewritten the pull request and added a lot, based on the various comment here.

The new proposal is here:
https://github.com/ossf/wg-best-practices-os-developers/pull/323

PR #154, opened 2023-05-25
Drop -Wl,-z,noexecheap

siddhesh:
noexecheap is a Gentoo downstream extension and is not available on any other distributions or in upstream linkers.


rcseacord (approved):
lgtm

thomasnyman (approved):
lgtm

PR #156, opened 2023-05-26
Update Concise-Guide-for-Evaluating-Open-Source-Software.md

carolinacanales:
Previous link was not working, added the right one: https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md


gkunz (approved):
LGTM

david-a-wheeler:
Can we modify this URL so the reader doesn't need to see all the GitHub preamble?

It looks like the trick of adding "#readme" isn't working.  But using this URL instead means that user won't need to see a lot of unnecessary header:

> https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#guide-to-implementing-a-coordinated-vulnerability-disclosure-process-for-open-source-projects

gkunz:
I see your point - at the same time, adding the reference to the heading increases the chances of breaking the link due to any sort of renaming of the heading... And renaming caused this issue in the first place.

david-a-wheeler:
> adding the reference to the heading increases the chances of breaking the link due to any sort of renaming of the heading... And renaming caused this issue in the first place.

If the anchor name link changes (the material after '#'), the user will still get to the page, just to the top instead of the intended header. That still seems like a win, you're "no worse off".

I think the best practices WG will soonish have a "better publishing process". If it seems useful, we can replicate it for any other group that desires it.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -32,7 +32,7 @@ As a software developer, before using open source software (OSS) dependencies or
    2. Is its interface/API designed to be easy to use securely (e.g., if the interface implements a language, does it support parameterized queries)?
    3. Is there guidance on how to use it securely?
 
-6. **Are there instructions on how to report vulnerabilities?** See the [Guide to implementing a coordinated vulnerability disclosure process for open source projects](https://github.com/ossf/oss-vulnerability-guide/blob/main/guide.md#readme) for guidance to OSS projects.,
+6. **Are there instructions on how to report vulnerabilities?** See the [Guide to implementing a coordinated vulnerability disclosure process for open source projects](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md) for guidance to OSS projects.,
```
gkunz, line 35: @carolinacanales:

```suggestion
6. **Are there instructions on how to report vulnerabilities?** See the [Guide to implementing a coordinated vulnerability disclosure process for open source projects](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#guide-to-implementing-a-coordinated-vulnerability-disclosure-process-for-open-source-projects) for guidance to OSS projects.
```

gkunz (changes requested):

Issue #160, opened 2023-06-06
Modify CSS of generated pages


david-a-wheeler:
We need to modify the CSS for the generated pages under /docs to "look like OpenSSF".

This will probably involve modifying assets/css/style.scss as well as some other pages.

Eventually we should be able able to import a common CSS so it could be reused by different sites.

jennbly:
Great idea! Yes, first draft of the OpenSSF brand guidelines with the OpenSSF logo/colors/fonts are available here: https://openssf.org/brand-guidelines/

hythloda:
I submitted a request with the web design team for the CSS. I spent a while trying to pull it from the Wordpress site to no avail. Will share when they get back to me.

ctcpip:
is there a good place to bring up a11y issues with the brand guidelines?  both the fonts are problematic for accessibility:

**Cairo:**
![Screenshot 2023-06-07 at 2 55 15 PM](https://github.com/ossf/wg-best-practices-os-developers/assets/15326526/18ee9aed-c6f4-4cc9-8b7e-348037ed419a)

**Teko:**
![Screenshot 2023-06-07 at 2 55 08 PM](https://github.com/ossf/wg-best-practices-os-developers/assets/15326526/d9d4ad79-e4cd-4c8b-881c-e920f458435f)

jennbly:
Could you elaborate further on what the concern is? This is something we can bring up with our design and web teams to investigate.

ctcpip:
@jennbly the legibility and readability of those fonts is not great -- the screenshots there are meant to demonstrate that.  using a better font would make OpenSSF web content more friendly for people with visual and cognitive disabilities.

ctcpip:
I'll note that the OpenSSF website uses a different font (Poppins) -- which is more legible and readable but still leaves much to be desired:

![Screenshot 2023-06-07 at 5 17 25 PM](https://github.com/ossf/wg-best-practices-os-developers/assets/15326526/ddc71900-8bda-4f70-a8a6-c3abe0925575)

although the `18px` font size on the website helps a lot

jennbly:
Agreed accessibility is very important. Thank you for the additional info. We will have a discussion with our design and web teams to determine the next steps forward.

ctcpip:
[...]

PR #168, opened 2023-06-07
Use OpenSSF fonts

david-a-wheeler:
I'm not sure the URLs are correct here, we may need to experiment. This code means we serve our fonts.


david-a-wheeler:
If anyone has experience with tweaking fonts like this, please let me know. I'm trying to switch to the "basics".

david-a-wheeler:
This is a *small* step towards implementing [issue 160](https://github.com/ossf/wg-best-practices-os-developers/issues/160), that is, to implement the draft [OpenSSF branding guidelines](https://openssf.org/brand-guidelines/) by using the current proposed fonts.

NB: There appear to be some questions about these fonts, perhaps we should wait until that's examined.

Any comments? Esp: @hythloda @SecurityCRob @jennbly ?

david-a-wheeler:
I think we should merge this & tweak if necessary to change the fonts. Even if we decide to use another font, it's at least a step towards configuring the generated web pages to look like we want them to look.

jennbly:
I'd agreed to go ahead and move forward with the proposed fonts as a good start in the right direction. And then once we have investigated the concerns and solutions with the design & web teams, if changes are needed, we can adjust accordingly.

david-a-wheeler:
Ok, per recommendation I'll merge this and we can see what it looks like.

My experience with CSS is that you often have to twiddle things to "get it right". Eventually we'll probably want one OpenSSF-wide repo that lets us update the CSS for all, but for now, let's get it working.

david-a-wheeler:
You can currently see the results here:

https://bp.openssf.org/Concise-Guide-for-Developing-More-Secure-Software.html

Now that I can see it, Ihe lack of serifs is unfortunate for when you have a lot of text. Are we sure we want a sans-serif font for this case?? I also think the heading font is excessively blocky, though that's just a personal preference. However, we have switched to the recommended format, and now we can be confident that we could switch again later.

Since there is no color on most of the guidebooks, there are no colors to change. Maybe I can call this "done"?

PR #171, opened 2023-06-08
Add an index.md to /docs of "released work"

david-a-wheeler:
Add an "index.md" file so that people can view
<https://best.openssf.org> and see all the *released* work from the OpenSSF Best Practices Working Group.

The goal is to make this more discoverable.

We *separately* need to create a "landing page" for developers, one that would cover all OpenSSF work (at least).
That is a *separate* idea, I currently expect that to be developed in `./developer.md`. But let's quickly get this "general list" added, so it'll be easy to find this info!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,0 +1,36 @@
+# OpenSSF Best Practices Working Group (WG)
+
+This is a list of materials (documents, services, and so on) released by the
+[Open Source Security Foundation (OpenSSF)](https;//openssf.org)
+[Best Practices Working Group (WG)].
+
+## Guides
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@SecurityCRob - as Best Practices WG lead, this okay? I think it's important to have a "basic list" of the WG available at the "published" website.

ctcpip (approved):

SecurityCRob:
Yes, I agree with this approach.

Cheers,

CRob
Director of Security Communications
Intel Product Assurance and Security

PR #172, opened 2023-06-09
Improve published metadata

david-a-wheeler:
This modifies `_config.yml` so that the generated pages on <https://best.openssf.org> will have more accurate metadata.


david-a-wheeler:
This is merely some config changes for the experiment in webserving, so I'm going to merge.

PR #173, opened 2023-06-09
Add favicons

david-a-wheeler:
To make things easy, I'm just copying the favicon invocations from the main openssf.org site. I'm also just referencing the files, instead of copying them in.


david-a-wheeler:
This just reproduces use of the favicons we already uses.

PR #191, opened 2023-06-29
Create documentation for "Simplest Possible Process" (SPP)

david-a-wheeler:


david-a-wheeler:
@SecurityCRob - here's a first edition. I propose merging and then iterating as needed.

SecurityCRob (approved):
Thanks for putting this together so quickly, David!

PR #213, opened 2023-08-10
Additional repository operational recommendations 

caabernathy:
Bundled a few changes here based on previous feedback:

1.  There was an ask to add language in the operational section around what one should do when taking a repo public - basically run through the repository configuration checks. So added a one-liner for that as well as checks that should also take place when a repo is transferred into an org.
2. Based on (1) it made sense to change the title in one of the headers from "Repository configuration" to "Repository". This is mainly because some items on that list are not really about configuration - ex: keeping a repository up to date or Scorecard score above 7.
3. Lastly, there's been some confusion around employees and tried to use language to make this apply only to say companies.


SecurityCRob (approved):
lgtm +1

PR #215, opened 2023-08-15
Propose adding 2 new principles

david-a-wheeler:
Add one on design and one on implementation.
This also fixes a trailing whitespace.

I renumbered all the entries; this increases the risk of a merge conflict, but markdownlint will complain if I don't do it.

### DIFF
--- a/docs/SecureSoftwareGuidingPrinciples.md
+++ b/docs/SecureSoftwareGuidingPrinciples.md
@@ -3,8 +3,8 @@
 **As developers of software, we are committed to enhancing the security and transparency of the software supply chain by pledging the following for all software we produce, both proprietary and open source, whether embedded in a device, released on a standalone basis, or designed to operate as a service, with the goal of creating software that is secure by default:**
 
 1. To employ development practices that are in conformance with modern, industry-accepted secure development methods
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@jkmarz - what do you think?

jkmarz:
@david-a-wheeler I'm of two minds... first thought is, could these be considered part of "modern software development practices" and could be moved to implementation guide? Second thought is, if added as principles, would be stronger without "where practical" language --- OK to remove?

david-a-wheeler:
Ok to remove "where practical", I'll do that.

You could argue everything listed here are "modern software development practices". I think these points add specifics that are important while still staying high-level.

david-a-wheeler:
I'll note that a majority of the course on developing secure software can be summarized by these two new principles.

jkmarz (approved):

david-a-wheeler:
Thanks for the approval!

I believe this resolves the substantive comments from the best practices WG meeting today, where some wanted something about design and implementation. It also resolves the CI/CD errors (by fixing the markdown).

Since @jkmarz was the original proposer & approved this change, I'll just merge this. If that's in error, please revert.

PR #216, opened 2023-08-16
Reword hardening guide text about "outdated"

david-a-wheeler:
Fix misleading text.

The point is that some people use outdated versions of GCC, but as written it sounds like GCC is always outdated. This makes a few more tweaks for clarity.


david-a-wheeler:
@gkunz - seem reasonable?

david-a-wheeler:
@thomasnyman ?

gkunz (approved):
Wait, what? GCC isn't outdated? :-)

Thanks for catching this!

thomasnyman (approved):

PR #226, opened 2023-08-30
Update URLs (BP badge and courses)

david-a-wheeler:
Update the URLs to be current.


ctcpip (approved):

david-a-wheeler:
Thanks! This simply fixes some URLs, so I'll just merge this.

There are some linting errors. We should address those, but it'd be confusing to address them in this PR, so I propose handling those separately.

PR #227, opened 2023-09-01
Update README.md

gkunz:
This is merely a cosmetic change, but I find it easier to read the table when the columns are left-justified and when the links to the individual guides are visually separated. Just based on my struggles trying to find the right guide in the Readme.


SecurityCRob (approved):
+1

Issue #202, opened 2023-07-18
Suggestions on Concise Guide for Evaluating Open Source Software


AndrewX192:
I recently came across the [Concise Guide for Evaluating Open Source Software](https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software)! Really glad to see people talking about this topic. In my past experience I've seen a few other areas that I believe are important to this evaluation. Happy to open a pull request but thought I'd field some thoughts on the direction you'd like this page to take first, given that there's a competing goal to be concise :) Most of these are focused more around how software is to be used rather than the posture of the software on it's own as I've found those to be a bit more valuable than considering a piece of software at a particular point in time:

1) How do you intend to use the software? Is it a build or runtime dependency? Will the software be incorporated into a versioned deliverable to users or up for the user to update themselves? Pay attention to not only how you would update the software in your latest build particularly where your use case is a mobile or desktop application.
- In the case that you only need some functionality of the software it may be helpful to understand the code structure in detail and only utilize specific portions of the software (e.g. classes, packages, modules, etc.) to solve your problem rather than the entirety of the software.
- If you only need the software for a debug / pre-release or as part of the build process ensure it isn't being utilized as a runtime dependency
2) How will you become aware of updates to this software? Are there regular releases or only a mainline branch? How will you approach updating this software (every version, minor only, etc?)
- If you are a significant user/contributor of this of software you may want to consider pre-release (embargoed) notice of future updates you can be ready to update.
3) Should a vulnerability be discovered, what can you to stop immediate bleeding? Can you degrade features, or disable functionality entirely?
- Consider implementing configuration options or other means of disabling functionality supported by this software if possible.
- In the event that a user is responsible for updating the dependency outside of your system, consider warning mechanisms to alert the user of out of date dependencies.
- If the software is not distributed together, consider emitting information about the version of any dependencies at startup so that the user can easily understand what version(s) were picked up and whether a particular runtime instance might be vulnerable.
4) What happens if the software becomes unmaintained? Are there alternatives you could switch to? Could you take on maintenance of the open source project? 
- It may be valuable to attempt to spend some time understanding the software and making some basic modifications to prepare for the potential to need to need to maintain this software on your own.
5) What will you do if the software does not meet a requirement of your software (e.g. scale, durability, etc.)?
[...]

PR #230, opened 2023-09-05
Add dependency usage considerations

AndrewX192:
Previous discussion in #202.

* Added a new document, [Evaluating Usage Concerns](https://github.com/AndrewX192/wg-best-practices-os-developers/blob/software-usage-concerns/docs/Evaluating-Usage-Concerns.md) to cover potential issues that might arise during the use of a dependency. 
* Revised the Concise Guide for Evaluating Open Source Software to include a note about obtaining information about potential security updates.

### DIFF
--- a/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
+++ b/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
@@ -46,6 +46,8 @@ As a software developer, before using open source software (OSS) dependencies or
    6. Consider running all defined test cases to ensure the software passes them.
    7. See the [OpenSSF's list of security reviews](https://github.com/ossf/security-reviews/blob/main/Overview.md#readme).
 
+As you integrate a particular dependency into your project, consider how you can plan for the future changes to the lifecycle of the depdendency: [Evaluating Usage Considerations](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Evaluating-Usage-Concerns.md)
+
 Other resources you may wish to consider include:
 
 1. [The Tidelift guide to choosing packages well (February 2021)](https://tidelift.com/subscription/choosing-open-source-packages-well), Tidelift
--- a/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
+++ b/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
@@ -51,7 +51,7 @@ As a software developer, before using open source software (OSS) dependencies or
    6. Consider running all defined test cases to ensure the software passes them.
    7. See the [OpenSSF's list of security reviews](https://github.com/ossf/security-reviews/blob/main/Overview.md#readme).
 
-As you integrate a particular dependency into your project, consider how you can plan for the future changes to the lifecycle of the depdendency: [Evaluating Usage Considerations](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Evaluating-Usage-Concerns.md)
+As you integrate a particular dependency into your project, consider how you can plan for the future changes to the lifecycle of the depdendency. See OpenSSF's [Evaluating Usage Considerations](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Evaluating-Usage-Concerns.md)
 
 Other resources you may wish to consider include:
 
--- a/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
+++ b/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
@@ -34,10 +34,15 @@ As a software developer, before using open source software (OSS) dependencies or
    2. Is its interface/API designed to be easy to use securely (e.g., if the interface implements a language, does it support parameterized queries)?
    3. Is there guidance on how to use it securely?
 
-6. **Are there instructions on how to report vulnerabilities?** See the [Guide to implementing a coordinated vulnerability disclosure process for open source projects](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#guide-to-implementing-a-coordinated-vulnerability-disclosure-process-for-open-source-projects) for guidance to OSS projects.
-7. **Does it have significant use?** Software with many users or large users may be inappropriate for your use. However, widely used software is more likely to offer useful information on how to use it securely, and more people will care about its security. Check if a similar name is more popular - that could indicate a typosquatting attack.
-8. **What is the software’s license?** Licenses are technically not security, but licenses can have a significant impact on security and sustainability. Ensure every component has a license, that it’s a widely-used [OSI license](https://opensource.org/licenses) if it’s OSS, and that it’s consistent with your intended use. Projects that won’t provide clear license information are less likely to follow other good practices that lead to secure software.
-9. **What are the results of code evaluation?** Even a brief review of software (by you, someone you hire, or someone else), along with recent changes to it, can give you some insight. Here are things to consider:
+6. **How will you become aware of security announcements & updates?**
+   1. Are there regular releases or only a mainline branch? How will you approach updating this software (every version, minor only, etc?)
+   2. Is there a newsletter or some other announcement board where security alerts and mitigations are announced?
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
Hmm, there are a lot of proposals here. Probably best to tease them apart.

First, there's this new proposed item:

> 6. **How will you become aware of security announcements & updates?**
>    1. Are there regular releases or only a mainline branch? How will you approach updating this software (every version, minor only, etc?)
>    2. Is there a newsletter or some other announcement board where security alerts and mitigations are announced?
>    3. If you are a significant user/contributor of this of dependency you may want to consider requesting pre-release (embargoed) notice of future updates if available.

I agree with the underlying *goal* but I do *not* agree with the specific solution being proposed here. We *already* address the first two points (in a different way) and the third one is best not done. Please let me explain!

I think the time for this kind of advice has ended. In 1995 I would have written the same kind of text myself, it was a good idea. In 2023 I think this is *not* a good idea. An empty React application typically has over 1,000 dependencies, and that's *before* it has any useful functionality. The first two points no longer make sense to most software projects. Updating on "every version" is typically *impossible*. Trying to *manually* monitor "newletters" for thousands of dependencies for security alerts is completely impractical. At the current & future scales, automation is the *only* answer.

Also, pre-release embargoed notices have often become legally challenging to release; if you do it for one group, you have to do it for other groups, and they may include may be malicious parties (so you risk pre-releasing to malicious actors before blameless users receive them). They happen sometimes, but it's increasingly not worth trying to recommend. They're increasingly rare and it's not clear we *WANT* people to try to do them, so I don't think we want to be recommending looking for them.

So what we should be doing is recommending that people monitor their software dependencies using at least one automated tool (Dependabot, Gemnasium, etc.). But note that this isn't a criterion for evaluation, it's a criterion for developing and maintaining *our* software. We *already* say this, in Concise-Guide-for-Developing-More-Secure-Software.md. In particular, we say:

> **Monitor known vulnerabilities in your software’s direct & indirect dependencies**. E.g., enable basic scanning via GitHub's [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates) or GitLab [dependency scanning](https://docs.gitlab.com/ee/user/application_security/dependency_scanning/). Many other third party Software Composition Analysis (SCA) tools are also available. Quickly update vulnerable dependencies.

This PR also includes many other proposed changes & whole new document. I appreciate the effort shown! However, I'll comment on that separately.

david-a-wheeler:
There's also a proposed addition here:

> As you integrate a particular dependency into your project, consider how you can plan for the future changes to the lifecycle of the depdendency. See OpenSSF's [Evaluating Usage Considerations](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Evaluating-Usage-Concerns.md)

Obviously the misspelling of "depdendency" needs to be fixed, but that's easily dealt with. The real question is this new long "Evaluating Usage Concerns" doc. I'm going read that & make comments, but I need to read it first :-). Stay tuned...

david-a-wheeler:
I've decided it's easier to review by quoting the new doc.

> # Evaluating Usage Considerations
> 
> As a software developer, before using open source software (OSS) dependencies or tools, it is important to ensure that they will fit into the long-term lifecycle of our use-case. To evaluate a potential OSS dependency for long-term use-case fit, consider these questions (all tools or services listed are merely examples):

There are some really interesting ideas here! Thanks! Some of these seem to *not* be specific to reusing code - maybe they should be general points in the "Developing" guide?  So let's look at each point...

> 1. **How do you intend to use the software?** Is it a build or runtime dependency? Will the software be incorporated into a versioned deliverable to users or up for the user to update themselves?

There are *multiple* use cases. Runtime dependency, build-time dependency, test-time dependency, developer-time dependency. In general, you should only require the software when you need it. Many package managers have way to annotate "use this dependency on in these cases", but many developers fail to do this, leading to massive unnecessary dependencies. So they should determine when they need it, *AND* be sure to ONLY require it when it's needed (e.g., by specifying this to a package manager).

Also: This seems to apply to code developed by the project, not just reused code. Maybe this belongs in the Concise Guide for Developing Software?

>    1. Consider the full lifecycle of the update process, including deprecating or blocking old client software that has not updated, particularly where your use case is a mobile or desktop application.

Well... sometimes. Sadly, it's often *not* possible to update old client software, e.g., due to hardware limitations, conflicting software, etc. I wouldn't strongly recommend this. A *warning* is sensible!

>    2. In the case that you only need some functionality of the software it may be helpful to understand the code structure in detail and only utilize specific portions of the software (e.g. classes, packages, modules, etc.) to solve your problem rather than the entirety of the software.

I don't think this is helpful advice. Even if you understood the current code structure in detail, the next update will make your knowledge useless. Developers of reusable code should emphasize their API, and hopefully keep that stable. Stick with the API, anything else is unlikely to be something you can depend on.

If your goal is "extract the code fragments you want to use" that's even *worse*. Now you've created an unmaintained fork. If you want the code split, it's better to work with the upstream project to get *them* to split it up, then you can depend on just that part.

>    3. If you only need the software for a debug / pre-release or as part of the build process ensure it isn't being utilized as a runtime dependency.

Agreed! I think that's part of a broader recommendation that needs to be made clear and its own point. It *is* a good point though!

> 2. **Should a vulnerability be discovered, what can you to stop immediate bleeding?** Can you degrade features, or disable functionality entirely?
>    1. Consider implementing configuration options or other means of disabling functionality supported by this software if possible.

Again, this isn't specific to reusable software. This true for the software you develop *and* the software you reuse.

It's a great idea, though. The best practices badge has a "read-only" mode that can be quickly enabled and disables writing to the database. It provides partial functionality if a really serious vulnerability is found.

Recap: I think this needs to be broadened, but it's a great point.

>    2. In the event that a user is responsible for updating the dependency outside of your system, consider warning mechanisms to alert the user of out of date dependencies.

"Consider" might be okay. However, on some systems this is just "a day ending in the word day" and doesn't really provide helpful information. It might be more helpful if it only warned when there was a known vulnerability that mattered to the system. Is there an easy way to do this? I suspect warning on just a few especially bad cases (rather than "everything") is doable.

>    3. If the software is not distributed together, consider emitting information about the version of any dependencies at startup so that the user can easily understand what version(s) were picked up and whether a particular runtime instance might be vulnerable.

This sounds like it'd be a huge amount of work. Also, a lot of systems are being pre-packaged as a unit (e.g., as containers) - does this make this advice useless?

>    4. Having a means to build the dependency from source rather than using a pre-built version can be helpful in emergencies.

This sounds like a "nice to have, but too hard and few will care so no one will do it". We should emphasize widely-useful practical advice, not special cases rarely used. I'm not sure this is worth the squeeze.
 
> 3. **What happens if the software becomes unmaintained?** Are there alternatives you could switch to? Could you take on maintenance of the open source project?
>    1. It may be valuable to attempt to spend some time understanding the software and making some basic modifications to prepare for the potential to need to need to maintain this software on your own.

The problem is real. I don't think this solution will work in many cases. If you have to take over maintaining everything, that means disaster.

Here are some potential alternatives:

1. If there is a standard interface implemented by more than one implementation, prefer using the standard interface where possible (and if not, try to limit extensions).
2. Work to help make software sustainable if it's at risk of becoming unmaintained. E.g., work to move it to a foundation, help to get multiple organizations to back it, etc.

> 4. **What will you do if the software does not meet a requirement of your software (e.g. scale, durability, etc.)?**
>    1. Your use-cases may not be amendable to the maintainer's aspirations for the project. Will you need to fork the software? If so, how will you ensure that your fork will not devolve far from the upstream's development progress?

The problem is real, but forking is a *last* resort. The *first* step should be to see if there's a way to implement that requirement while being compatible with the project. E.g., improving the project software and/or its test suite, creating a plug-in system so you can add special requirement implementations, etc.

I think there are some good things here, but I think work is needed to make them broadly applicable. Thank you *so* *much* for raising these issues!

AndrewX192:
Several months have gone by since I opened this PR and unfortunately I am no longer able to contribute to Linux Foundation projects due to employment constraints. Seeing as there’s very little if anything to change I’ll go ahead and close out this PR and related issue as WONTFIX.

PR #231, opened 2023-09-07
♿️ add text to GitHub/GitLab refs

ctcpip:
also included:

- minor markdown fixes
- spelling fixes
- ~restores tooling alpha ordering which the group agreed to keep, but a PR was previously merged erroneously with that change in place~ this change was already merged in a different PR earlier today (#223) and no longer appears in the changes for this PR after I rebased from `main` 

note: this does not fix the fact that the non-link svgs are still linked by GitHub's md parser.  we won't be able to fix this until and unless we use a different image format (e.g. `gif`), which would give us some options for inserting the images without linking


torgo (approved):
lgtm

SecurityCRob:
+1

PR #233, opened 2023-09-13
Fix missing binutils version for rpath

### LINKED ISSUE #164: Compiler Hardening Guide: Missing Binutils version for `-rpath` and `--enable-new-dtags`
The `-Wl,-rpath` and `-Wl,--enable-new-dtags` options under *Discouraged Compiler Options* section are missing Binutils version information. The missing version information should be added for consistency.

thomasnyman:
Fixes #164.

Also removes `-Wl,--enable-new-dtags` from list of discouraged options as per @siddhesh's feedback:

> Looking at the _Discouraged Compiler Options_ section a bit closer though, ISTM that the mention of `-Wl,--enable-new-dtags` option is unnecessary. All it does is make the linker generate `DT_RUNPATH` instead of `DT_RPATH` when `-Wl,-rpath` is passed, it doesn't really do anything on its own. Could we just drop it?

Finally, adds an anchor span to the `-Wl,-rpath` description and hyperlink to it from the discouraged options table.


siddhesh:
LGTM.

PR #244, opened 2023-10-04
Update SecureSoftwareGuidingPrinciples.md

SecurityCRob:
updates following TAC feedback from 3Oct2023


docs/SecureSoftwareGuidingPrinciples.md
ware, line 1: Do we want to wait on 1.0 until we have GB sign-off?  Maybe 0.9?  Just in case GB wants changes?

ware, line 1: Do we want to use "throughout" instead of "through out"?

ware (commented):
Just a couple of small comments.  In general, very nice addition.

PR #253, opened 2023-10-11
Add note in table 1 about -Werror

david-a-wheeler:
We already talk about -Werror in more detail, but a quick note in table 1 supporting it would be wise.


david-a-wheeler:
@thomasnyman - I think this should be merged *before* the document version 1.0 is announced. The `-Werror` option is more controversial if this point isn't made clear, and some people might skip the prose that discusses it.

siddhesh:
Is `-Werror` use that controversial? More projects have been trying to do this lately by default (glibc and IIRC even binutils ship with `-Werror` on by default) and it seems like the main thing most developers struggle with is the correct strategy to disable `-Werror` where one needs to let a warning go through.  For example, using `-Wno-error=warning-type` is much better than `-Wno-warning-type` or a blanket `-Wno-error`.  We use pragmas in glibc to disable warnings at specific locations but that may be too heavy to add as a recommendation.

david-a-wheeler:
Yes, though I don't think "controversy" is the best term. This flag is helpful in some cases and not in others. `-Werror` should normally not be used when distributing source code, because it essentially makes it impossible to ever update the compiler. You *do* want `-Werror` enabled when you're modifying the source code, because now you're in a position to fix it.

thomasnyman:
I tweaked the statement on `-Werror` in the TL;DR section to provide some additional context and reworded the references  _"redistributors"_ to instead refer to _"when distributing source code"_ to avoid confusion between _source code redistributors_ and _binary redistributors_ (`-Werror` wouldn't be problematic for binaries, after all).

Similarly the in-depth description now mirrors the statement in TL;DR (with references) and I moved some of the text which is not related to `-Werror` back to the TL;DR, although into a separate paragraph to clarify that it applies to other options as well. 

Finally resolved the conflicts with the main branch.

david-a-wheeler:
@thomasnyman - good point. I modified "not in distribution" to "not in source distribution" to make it all consistent.

david-a-wheeler:
I *think* it's ready to merge now. I'd like to get many of these PRs merged if we can.

gkunz (approved):

david-a-wheeler:
Okay, I'm going to merge this before yet another merge conflict appears :-).

PR #259, opened 2023-10-13
Add note about system includes to hardening guide

david-a-wheeler:


david-a-wheeler:
@thomasnyman - what do you think? This is an attempt to respond (in part) to a thoughtful comment from Daniel Stenberg @bagder about system headers. By *default* system headers are ignored for warnings, but it's possible to include them (generally a bad idea). You also usually want to consider third party include files as system files (you don't control them, so you're unlikely to be able to do much about their warnings).

david-a-wheeler:
[siddhesh](https://github.com/siddhesh):

> However this also reduces the possibility of detecting a library/third-party bug that may not have been exposed elsewhere. Sure, it's a bug outside the application, but it still manifests in the application context, doesn't it?

It's generally not something the application developer can do anything about. Even worse, what's likely to happen is that the application developer would simply disable all the warning flags triggered by the system software, leading to the application software being *worse*.

They'd still be detected as long as the system or third party software developers enabled them when *they* compile things. If they don't, that's the problem.

david-a-wheeler:
There's a reason "ignoring system headers" is the default - application developers generally can't do anything about the problem.

bagder:
> I'm curious to know more about examples where system headers induced noise

They don't just "induce noise". With `-Werror` and friends they cause red CI jobs. They stand in the way of development of your project. Sure we can (and should?) fix our neighbors as well, but they might not agree, they may be slow, and most importantly **we might still want to be able to build with older versions of the 3rd party code**.

Example: in my projects I often stick to C89, while I also make the project able to *optionally* build with 3rd parties that might opt to use C99. Or the 3rd parties simply have not the same practice of using the strictest possible warning levels.

siddhesh:
Thanks for your responses, I empathize with those concerns. I'd be happy if alongside the use of `-isystem` and `-Wno-system-headers` as a way to silence warnings in headers included from third party libraries, we also point out the tradeoff, i.e. potentially hiding bugs that affect the application.

david-a-wheeler:
@siddhesh : That's a good point. I'll push an update to this PR to try to address this. Here's my proposed text:

>> There are trade-offs. Silencing warnings from system headers and third party libraries may hide vulnerabilities in them that affect the application.

[bagder](https://github.com/bagder) said:

> They don't just "induce noise". With -Werror and friends they cause red CI jobs. They stand in the way of development of your project. Sure we can (and should?) fix our neighbors as well, but they might not agree, they may be slow, and most importantly we might still want to be able to build with older versions of the 3rd party code.

Also a good point. I propose adding this as well:

>> On the other hand, *not* silencing them focuses efforts on issues that the developer typically cannot control, impede progress when using `-Werror` in CI jobs, and often make it difficult to support building with older versions of third party code.

david-a-wheeler:
I think adding the discussion about trade-offs makes this clearer. Hopefully we've resolved things. Seem okay?

siddhesh:
That reads well, thanks.

gkunz (approved):
This looks good to me. The new paragraph covers the reasons and relavant tradeoffs.

david-a-wheeler:
Thomas will fix the links (convert to references)

thomasnyman:
Made the inline links into references in footnotes and merging this as per the discussion in [C/C++ BP Call on 2023-10-25](https://docs.google.com/document/d/1UClGUnOSkOH_wab6Lx43KUdkaK37L8sbWJ_GPZvc1YY/edit#bookmark=id.d6uxl5fycds1).

Issue #250, opened 2023-10-11
Feedback before initial release of Compiler Options Hardening Guide for C and C++


thomasnyman:
Opening this issue for collecting feedback on the proposed [**Compiler Options Hardening Guide for C and C++**](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++) before its initial release.

Tagged release candidate for the guide can be found [here](https://github.com/ossf/wg-best-practices-os-developers/blob/compiler-hardening-guide-rc1/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C%2B%2B.md).

tcullum-rh:
Hi,

This guide is quite comprehensive! My only current suggestion is that as a developer, I would really appreciate more info about this:

> 
Those who build programs for production may choose to omit some options that hurt performance if the program only processes trusted data, but remember that it’s not helpful to deploy programs that that are insecure and rapidly do the wrong thing. Existing programs may need to be modified over time to work with some of these options.

It would be helpful if for each option (or only the relevant options), a statement was made as to what the performance or other drawbacks are and any relevant workarounds.

thomasnyman:
> It would be helpful if for each option (or only the relevant options), a statement was made as to what the performance or other drawbacks are and any relevant workarounds.

Thank you for the feedback @tcullum-rh. 

They way in which we have tried to accommodate this in the current version of the guide are the **Performance implications** and **When not to use?** sections in the detailed descriptions for each option (excluding the Warning options, as they do not impact the produced binary directly).

For example, for [-fstack-protector-strong](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C%2B%2B.md#performance-implications-3):

>
> Stack protector supports three different heuristics that are used to determine which functions are instrumented with run-time checks during compilation:
>
> -fstack-protector-strong: instrument any function that takes the address of any of its local variables on the right-hand-side of an assignment or as part of a function argument allocates a local array, regardless of type or length
allocates a local struct or union which contains an array, regardless of the type of length of the array
has explicit local register variables
> -fstack-protector: instrument functions that call alloca() or allocate character arrays of n bytes or more in size . The threshold for instrumentation is adjustable via the --param=ssp-buffer-size=n option (default: 8 bytes).
[...]

PR #261, opened 2023-10-17
When compiling compilers, make their defaults secure

david-a-wheeler:


david-a-wheeler:
@eslerm @thomasnyman - comments?

kees:
Perhaps refer to Ubuntu's compiler flag wiki page, as this is what Ubuntu already does. (I'm not aware of any other distro that enables hardening options in their compiler, unfortunately.)
https://wiki.ubuntu.com/ToolChain/CompilerFlags

setharnold:
> The Ubuntu compiler flag wiki page is out of date when it comes to PIE (flags are not set, but compiler is built with `--enable-default-pie`). I'll be addressing this.

Depending upon the context of the discussion, either one might be appropriate. A developer looking for the 'best' options to use in their build might want the `-fPIE` or `-fPIC` options but someone packaging GCC might want the `--enable-default-pie` compiler option.

siddhesh:
> Perhaps refer to Ubuntu's compiler flag wiki page, as this is what Ubuntu already does. (I'm not aware of any other distro that enables hardening options in their compiler, unfortunately.) https://wiki.ubuntu.com/ToolChain/CompilerFlags

I believe Gentoo also defaults to enabling many hardening features in their toolchain.

eslerm:
clang sets `--enable-default-pie` as [a default now](https://reviews.llvm.org/D120305):
> CMake ``-DCLANG_DEFAULT_PIE_ON_LINUX=ON`` is now the default. This is used by
> linux-gnu systems to decide whether ``-fPIE -pie`` is the default (instead of 
> ``-fno-pic -no-pie``). This matches GCC installations on many Linux distros.
> Note: linux-android and linux-musl always default to ``-fPIE -pie``, ignoring
>  this variable. ``-DCLANG_DEFAULT_PIE_ON_LINUX`` will be removed in the future.

However, gcc itself [does not](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103398).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -63,6 +63,10 @@ Some organizations require selecting hardening rules. For example, the US govern
 
 [^CMU2018]: Carnegie Mellon University (CMU), [Top 10 Secure Coding Practices](https://wiki.sei.cmu.edu/confluence/display/seccode/Top+10+Secure+Coding+Practices), SEI CERT Coding Standards Wiki, 2018-05-02.
 
+**What should you do when compiling compilers?**
+
+If you are compiling a C/C++ compiler, where practical make the generated compiler's default options the *secure* options (as documented here). For example, when compiling GCC, use `--enable-default-pie` (which makes `-fPIE` and `-pie` the defaults) and when compiling clang on Linux systems set `CLANG_DEFAULT_PIE_ON_LINUX` (which has a similar effect).
```
eslerm, line 68: to clarify configurations vs flags:

"when compiling GCC, use the configuration `--enable-default-pie` (which enables the flags `-fPIE` and `-pie` by default)"

eslerm (changes requested):

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Done!

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -63,6 +63,10 @@ Some organizations require selecting hardening rules. For example, the US govern
 
 [^CMU2018]: Carnegie Mellon University (CMU), [Top 10 Secure Coding Practices](https://wiki.sei.cmu.edu/confluence/display/seccode/Top+10+Secure+Coding+Practices), SEI CERT Coding Standards Wiki, 2018-05-02.
 
+**What should you do when compiling compilers?**
+
+If you are compiling a C/C++ compiler, where practical make the generated compiler's default options the *secure* options (as documented here). For example, when compiling GCC, use `--enable-default-pie` (which enables the flags `-fPIE` and `-pie` by default when using the generated compiler executable) and when compiling clang on Linux systems set `CLANG_DEFAULT_PIE_ON_LINUX` (which has a similar effect).
```
gkunz, line 68: nit-pick: when reading the text in the parenthesis, I initially expect the "here" to be a link instead of referring to this paragraph.

```suggestion
If you are compiling a C/C++ compiler, where practical make the generated compiler's default options the *secure* options. For example, when compiling GCC, use `--enable-default-pie` (which enables the flags `-fPIE` and `-pie` by default when using the generated compiler executable) and when compiling clang on Linux systems set `CLANG_DEFAULT_PIE_ON_LINUX` (which has a similar effect).
```

gkunz (commented):
LGTM expect for the small nit-pick.

thomasnyman:
I can understand the rationale for including this, but I'm not so sure about whether the introductory section (which is already rather long) is the best place for this. I'm mindful of the feedback we have received regarding the [amount of prose](https://github.com/ossf/wg-best-practices-os-developers/issues/250#issuecomment-1761772847). 

I'm also not a big fan of jumping straight into specific options in this section (apart from the TLDR). 

Would this be a better fit as either a separate section or appendix with the options in a table for consistency?

david-a-wheeler:
Let's move it later into its own section. Most people don't compile their own compilers.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thesamesam: Clang would like to deprecate that CMake option and is instead pushing people towards Clang configuration files.

You can see what we do in Gentoo at https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-devel/clang-common/clang-common-18.0.0_pre20231025.ebuild?id=b162a825ab14ae965ca012599f5efe9283ff7c36#n131.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -63,6 +63,10 @@ Some organizations require selecting hardening rules. For example, the US govern
 
 [^CMU2018]: Carnegie Mellon University (CMU), [Top 10 Secure Coding Practices](https://wiki.sei.cmu.edu/confluence/display/seccode/Top+10+Secure+Coding+Practices), SEI CERT Coding Standards Wiki, 2018-05-02.
 
+**What should you do when compiling compilers?**
```
thesamesam, line 66: For GCC, you probably want `--enable-default-ssp` which uses `-fstack-protector-strong`.

thesamesam:
> > Perhaps refer to Ubuntu's compiler flag wiki page, as this is what Ubuntu already does. (I'm not aware of any other distro that enables hardening options in their compiler, unfortunately.) [wiki.ubuntu.com/ToolChain/CompilerFlags](https://wiki.ubuntu.com/ToolChain/CompilerFlags)
> 
> I believe Gentoo also defaults to enabling many hardening features in their toolchain.

You can see what we do at https://wiki.gentoo.org/wiki/Hardened/Toolchain#Changes. We don't set any of that via flags-per-package, instead via either patches or configure arguments to GCC or Binutils. For Clang, we use config files in `/etc/clang` as upstream are trying to move towards that rather than adding many CMake configuration options for stuff like this (we barely got them to add PIE).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Okay, I'll do that.

david-a-wheeler:
I think this tweaked version covers the gist: If you control compiling the compiler, make the defaults the secure ones.

That said, many people aren't recompiling their compilers, so we need to give them guidance for their cases. If they include lots of options in their builds, the code is likely to work with those options elsewhere too :-).

thesamesam (approved):
We should talk about doing more here in future, but the guidance here is solid and there's no reason not to just add more stuff later.

Thanks!

gkunz (approved):

david-a-wheeler:
Ok. I'll merge this now, we can add more later. Thank you everyone!

PR #262, opened 2023-10-18
Add -Wimplicit-fallthrough

david-a-wheeler:
This also documents how to mark *intentional* implicit fallthrough. Developers might be tempted to *not* use this warning if it's not clear how to handle this case.

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -117,7 +117,7 @@ Table 1: Recommended compiler options that enable strictly compile-time checks.
 | [`-Wformat=2`](#-Wformat=2)                                                   | GCC 2.95.3<br/>Clang 4.0 | Enable additional format function warnings                                          |
 | [`-Wconversion`](#-Wconversion)<br/>[`-Wsign-conversion`](#-Wsign-conversion) | GCC 2.95.3<br/>Clang 4.0 | Enable implicit conversion warnings                                                 |
 | [`-Wtrampolines`](#-Wtrampolines)                                             |         GCC 4.3          | Enable warnings about trampolines that require executable stacks                    |
-| [`-Wimplicit-fallthrough`](#-Wimplicit-fallthrough)                           |         GCC<br>Clang   | Warn when a switch case falls through                                                 |
+| [`-Wimplicit-fallthrough`](#-Wimplicit-fallthrough)                           |         GCC 7.1.0<br>Clang   | Warn when a switch case falls through                                                 |
 | [`-Werror`](#-Werror)<br/>[`-Werror=`*`<warning-flag>`*](#-Werror-flag)       | GCC 2.95.3<br/>Clang 2.6 | Make compiler warnings into errors                                                  |
 
 Table 2: Recommended compiler options that enable run-time protection mechanisms.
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@kees @thomasnyman - this PR proposes to add `-Wimplicit-fallthrough` as a compiler flag. I think it's a *great* idea.

I included documentation on how to portably *suppress* the warning in the rare cases when the fallthrough is intentional. If developers don't know how to portably annotate this, they might not add the flag.

Comments? Problems? Does this seem okay?

kees:
I don't think it's right to say there's no standard way to do this with C, since `__attribute__((__fallthrough__))` is how to express it. Leaving the kernel example for generating the macro isna good idea, though.

david-a-wheeler:
@kees : I tweaked the text, hopefully that clarifies things per your question.

thomasnyman:
Really like this addition. Especially with the concrete recommendations for dealing with the caveats differing attributes etc.

Proposed some fixes to the references for consistency in b274a223aa6ec43f846619e5f49c3b74cb93ae4d and minor rewording to the last sentence in eb8c4072fe204f45f056a9546bc4e4f4c9b3a926.

david-a-wheeler:
Thomas will check compiler version numbers and add that information.

david-a-wheeler:
@thomasnyman - please add the version number info when you get a chance. I'd like to see a lot of merges of pending changes, there's at least 1 issue I can't create a PR for, because it would lead to merge conflicts with existing PRs.

gkunz (approved):
LGTM

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -116,6 +117,7 @@ Table 1: Recommended compiler options that enable strictly compile-time checks.
 | [`-Wformat=2`](#-Wformat=2)                                                   | GCC 2.95.3<br/>Clang 4.0 | Enable additional format function warnings                                          |
 | [`-Wconversion`](#-Wconversion)<br/>[`-Wsign-conversion`](#-Wsign-conversion) | GCC 2.95.3<br/>Clang 4.0 | Enable implicit conversion warnings                                                 |
 | [`-Wtrampolines`](#-Wtrampolines)                                             |         GCC 4.3          | Enable warnings about trampolines that require executable stacks                    |
+| [`-Wimplicit-fallthrough`](#-Wimplicit-fallthrough)                           |         GCC<br>Clang   | Warn when a switch case falls through                                                 |
```
gkunz, line 120: GCC should support -Wimplicit-fallthrough since 7.1.0 [1]. I couldn't find a reference for Clang yet.

```suggestion
| [`-Wimplicit-fallthrough`](#-Wimplicit-fallthrough)                           |         GCC 7.1.0<br>Clang   | Warn when a switch case falls through                                                 |
```

[1] https://gcc.gnu.org/gcc-7/changes.html

david-a-wheeler: Clang appeared to have it by 2012: https://discourse.llvm.org/t/should-wimplicit-fallthrough-require-c-11/25199/13
... that doesn't answer the question, but perhaps it will help find the answer.

david-a-wheeler:
We're spending more time fighting merge conflicts than solving problems. All approved. I've merged it. If we ever find a specific Clang version, we can add it. However, we know it was in clang in 2012 (11 years ago), so I don't think documenting the specific clang version is *that* important, and I don't want the perfect to be the enemy of the good. Adding this option ends up modifying a line that just *begs* for merge conflicts. Now that we've added it, adding the clang version later should be easily handled as a separate PR.

PR #268, opened 2023-10-25
New document for gcc and clang annotation recommendations

siddhesh:
Add a document with recommendation for clang and GCC attributes.  This is a first draft and needs more information to be added to better describe the implications of each attribute.

The document is currently also missing the standard C attribute names for many of these attributes, which could potentially make them even more portable.


docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Clang and GCC Attributes at a glance
+
+Both GCC and Clang recognize the `__attribute__` keyword to annotate source code.  Both compilers also provide a `__has_attribute()` macro function that returns 1 if the attribute name passed to it is supported and 0 otherwise.  For example `__has_attribute(malloc)` would return 1 in the latest GCC and Clang.  The full syntax description for the `__attribute__` keyword is available in the GCC documentation[^1].  Since all of the attributes recommended in this document are function attributes, here's a quick primer on how they are used.
```
GabrielDosReis, line 7: I support the general idea of encouraging annotation that helps improve safety and security.  However, I have concerns about the syntax recommendations.  Where possible, we should encourage standard notation.  For instance, `[[fallthroug]]` is a standard C++ notation, now also available in C23.  I think a recommendation along the lines of "use these standard notations where available, or fall back these non-standard notations" would be a better recommendation.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -0,0 +1,42 @@
...
+
+When declaring functions, attributes may be added to the function at the end of the declaration, like so:
+
+```
+extern void *custom_allocator (size_t sz) __attribute__ ((malloc)) __attribute__ ((alloc_size (1)));
+```
```
GabrielDosReis, line 13: These are also available as `[[gnu::]]` attributes, right?

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
siddhesh: Ack, I wanted to put this work out for early feedback. I can modify examples (and indeed, all of the text) to use the C/C++ standard notation.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
siddhesh: Yes.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
david-a-wheeler: One challenge is that code often needs to be compilable with many existing and older compilers. So syntax from new specs often takes a while before people can use it in code. We should mention a spec where it exists, but also note an alternative when that option isn't available to a significant number of cases.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -0,0 +1,42 @@
...
+| `fd_arg_write(N)`                                          | GCC 13 | Function | Argument N is an open file descriptor that can be written to.                   |
+| `noreturn`                                          | GCC<br/>Clang | Function | The function does not return.                   |
+| `tainted_args`                                          | GCC 12<br/>Clang TBD | Function or function pointer | Function needs sanitization of its arguments. Used by -fanalyzer=taint                  |
+
+## Performance considerations
+
+Attributes influence not only diagnostics generated by the compiler but also the resultant code. As a result, annotating code with attributes will have an impact on performance, although the impact may go either way.  The annotation may allow compilers to add more traps for additional security and be conservative about some optimizations, thus impacting performance of output code. At the same time however, it may allow compilers to make some aggressive optimization decisions, resulting in generation of smaller and faster running code and often, better code layout.
+
+[^1]: Attribute Syntax
+<https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html>
```
gkunz, line 42: Nit-pick: I started by placing both lines in one to make sure that Jekyll renders the footnote correctly - and while on it, I added some additional text to make the footnote consistent with other references to the GCC manual.

```suggestion
[^1]: GCC team, [Attribute Syntax](https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html), GCC Manual, 2023-11-03.
```

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
GabrielDosReis: > So syntax from new specs often takes a while before people can use it in code. We should mention a spec where it exists, but also note an alternative when that option isn't available to a significant number of cases.

I do agree; it is worth noting that the standard notations have been available in at least the 5 major releases or GCC (which is more than supported/maintained LTS releases), and many more in GCC.   In our guidelines, I think we should strive to guide users to the standard notations first, where available, and fallback to non-standard notations only when no alternatives exist.

david-a-wheeler: If I understand you correctly I agree 

Let's document and recommend and prefer standard forms. Let's also document the alternatives in the unfortunate cases where that won't work for some reason.

david-a-wheeler:
@siddhesh : Will you make adjustments per the reviews above?

Does the group believe that if these adjustments were made, this should be accepted?

david-a-wheeler:
Also: I'd be more directive about standards-based approaches. "In general, prefer approaches that are defined by standards and supported by multiple compilers, e.g., in C++ prefer `[[fallthrough]]` over a compiler-specific attribute." Or something like that.

siddhesh:
> @siddhesh : Will you make adjustments per the reviews above?

Yes, I just haven't found a block of time to knock them off 😞 . I'll try to do it tonight or tomorrow so that we can discuss it during the call tomorrow if needed.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
siddhesh: Thanks, fixed.

PR #269, opened 2023-10-25
Werror note

david-a-wheeler:


thomasnyman:
This PR seems to be a duplicate of #253. Closing this in favor of #253 which has the related discussion and has newer commits.

PR #270, opened 2023-10-27
Add simple test of C/C++ hardening flags

david-a-wheeler:


david-a-wheeler:
Add a simple test of the C/C++ hardening flags; just run "make" (presuming this is GNU make).

david-a-wheeler:
I should note that the tests currently fail on at my Apple MacOS system:

Apple clang version 14.0.3 (clang-1403.0.22.14.1) fails with:

~~~~
make CC=clang
clang -O2 -Wall -Wformat=2 -Wconversion -Wtrampolines -Werror  -D_FORTIFY_SOURCE=3  -D_GLIBCXX_ASSERTIONS  -fstack-clash-protection -fstack-protector-strong  -Wl,-z,nodlopen -Wl,-z,noexecstack  -Wl,-z,relro -Wl,-z,now  -fPIE -pie -fPIC -shared     demo.c   -o demo
clang: error: argument unused during compilation: '-fstack-clash-protection' [-Werror,-Wunused-command-line-argument]
clang: error: argument unused during compilation: '-pie' [-Werror,-Wunused-command-line-argument]
~~~~

I think clang wants `-fpie` not `-pie`, per:
https://clang.llvm.org/docs/ClangCommandLineReference.html
... and it doesn't support `-Wtrampolines`. We should probably fix the former and discuss the latter.

~~~~
make CC=/usr/local/Cellar/gcc/13.2.0/bin/gcc-13
/usr/local/Cellar/gcc/13.2.0/bin/gcc-13 -O2 -Wall -Wformat=2 -Wconversion -Wtrampolines -Werror  -D_FORTIFY_SOURCE=3  -D_GLIBCXX_ASSERTIONS  -fstack-clash-protection -fstack-protector-strong  -Wl,-z,nodlopen -Wl,-z,noexecstack  -Wl,-z,relro -Wl,-z,now  -fPIE -pie -fPIC -shared     demo.c   -o demo
ld: unknown option: -z
~~~~

I think this is a configuration problem on my system. I think gcc is trying to use the Apple MacOS ld command at /usr/bin/ld, which is a hobbled mess, instead of GNU ld. But I suspect other users will see this, so we probably ought to discuss the issue of different loaders (ld) supporting different options.

gkunz:
Hi @david-a-wheeler, I really like the idea of adding a test - even though (or just because?) it surfaces some issues with the current set of options.

I can confirm your findings on my Mac (and I assume we basically the same version of MacOS 14.1).

Moreover, after getting too annoyed by my Mac, I ran the demo on an older Ubuntu 22.04.3 LTS with the following result:

    cc -O2 -Wall -Wformat=2 -Wconversion -Wtrampolines -Werror  -D_FORTIFY_SOURCE=3  -D_GLIBCXX_ASSERTIONS  -fstack-clash-protection -fstack-protector-strong  -Wl,-z,nodlopen -Wl,-z,noexecstack  -Wl,-z,relro -Wl,-z,now  -fPIE -pie -fPIC -shared     demo.c   -o demo
    <command-line>: error: "_FORTIFY_SOURCE" redefined [-Werror]
    <built-in>: note: this is the location of the previous definition

It seems that gcc on Ubuntu has predefined FORTIFY_SOURCE as something else but 3 and therefore it complains about redefining it. One solution is to undefine FORTIFY_SOURCE first and then setting it again.

    cc -O2 -Wall -Wformat=2 -Wconversion -Wtrampolines -Werror -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3  -D_GLIBCXX_ASSERTIONS  -fstack-clash-protection -fstack-protector-strong  -Wl,-z,nodlopen -Wl,-z,noexecstack  -Wl,-z,relro -Wl,-z,now  -fPIE -pie -fPIC -shared     demo.c   -o demo

Other findings?

david-a-wheeler:
> One solution is to undefine FORTIFY_SOURCE first and then setting it again.

That sounds like a good solution. We should propose it and explain why (in a separate PR). Will you write that up @gkunz ?

gkunz:
@david-a-wheeler yes

david-a-wheeler:
All agreed to adding this simple test, so I'm merging it. The test has revealed some problems - they need to be handled by other PRs. I plan to create a separate PR to turn `-pie` into `-fpie` but obviously that's not a *test*.

PR #276, opened 2023-11-03
C/C++ hardening flags: -pie should be -fpie

david-a-wheeler:
Testing revealed a bug, -pie should be -fpie. This fixes the bug.


david-a-wheeler:
@gkunz @thomasnyman - thoughts?

thomasnyman:
What is the test case? 

I believe for both GCC and Clang `-pie` is correct when the intention is to produce a position-independent executable, but it is a [link option](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-pie) so it only gets used when producing the final executable binary. 

The `-fpie` and `-fPIE` options I believe are synonymous, so the suggested change here ends up being redundant.

However, the interaction between `-fPIE -pie` and `-fPIC -shared` doesn't seem desirable here. Combining these sets of options makes producing a shared library take precedence. 

Would it make sense to drop `-fPIC -shared` in the TL;DR; and rather refer to reader to the detailed description for these options if they are building a shared library instead?

david-a-wheeler:
> What is the test case.

In the main branch of this repo, on a MacOS (Intel) computer:

~~~~
cd wg-best-practices-os-developers/docs/Compiler-Hardening-Guides
make
~~~~

Output is:
~~~~
cc -O2 -Wall -Wformat=2 -Wconversion -Wtrampolines -Wimplicit-fallthrough  -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3  -D_GLIBCXX_ASSERTIONS  -fstack-clash-protection -fstack-protector-strong  -Wl,-z,nodlopen -Wl,-z,noexecstack  -Wl,-z,relro -Wl,-z,now  -fPIE -pie -fPIC -shared     demo.c   -o demo
clang: warning: argument unused during compilation: '-fstack-clash-protection' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
warning: unknown warning option '-Wtrampolines' [-Wunknown-warning-option]
1 warning generated.
ld: unknown option: -z
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [demo] Error 1
~~~~

Running `cc --version` shows:
~~~~
cc --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: x86_64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
~~~~

david-a-wheeler:
Building shared libraries is pretty common. We should help people maximize the security of shared libraries as well as executables that aren't. I suspect that there needs to be some text to write that out, but I don't have time at the moment. Could you take a crack at it?

thomasnyman:
In 26ce885b617eafda959310bb34a379f9f1372ddf I separated the tl;dr; code listing into a separate listing from executables that uses `-fPIE -pie` and a separate one for shared libraries using `-fPIC -shared`.  I believe `-Wl,-z,relro` and `-Wl,-z,now` only affect produced executables so I omitted those the from the latter.

The `ld: unknown option: -z` error on MacOS is because MacOS `dyld` does not support the Binutils `ld` `-z`flag. I believe the `-z` flag options are all specific to ELF binaries.

david-a-wheeler:
This seems problematic. All shared libraries are also executables.

Also, it's pretty common for a build process to generate *both* shared libraries and direct executables - I think it'd be better to have a *common* set in one place, then *add* options for various special cases. This may end up being a new table. Basically, we may need a long "always" set, then a table showing "In this case... add these option(s)".

thomasnyman:
> This seems problematic. All shared libraries are also executables.

The [GCC](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#index-fpie) documentation refers to the artifacts produced by `-fPIE` as "_executables_" though (as opposed to code suitable for dynamic linking produced with `-fPIC`)

> Also, it's pretty common for a build process to generate _both_ shared libraries and direct executables - I think it'd be better to have a _common_ set in one place, then _add_ options for various special cases. This may end up being a new table. Basically, we may need a long "always" set, then a table showing "In this case... add these option(s)".


Given that we have (for now) also decided to list architecture-specific options in the tl;dr, section, maybe we could merge all of these into some kind of "_situational options_" table. For instance:

Situational options:
| for executables | for shared libraries | for x86_64 | for aarch64 |
|----------------|--------------------|------------|--------------|
| `-fPIE -pie`      |  `-fPIC -shared`     | `-fcf-protection=full` | `-mbranch-protection=standard`

thomasnyman:
Come to think of it, it would be quite nice to have some embedded Javascript that would let you pick the target artifact (application or library), the target OS and the target architecture and show you the relevant options for the selected combination.

Not sure how to maintain the corresponding lists of options in a sensible way though. Possible we could have some way of tagging them in the tables (even if the tags would be invisible in the rendered output).

thomasnyman:
Rebased this PR to correspond to the current main branch and updated this with the proposal based on a table for situational options.

thomasnyman:
Pivoted table as proposed in #300 and rebased against current `main`.

david-a-wheeler:
Wow, that was messy. I prefer avoiding rebasing a public branch, but I've resynchonized.

I really like the format of the pivoted table. I think the left-hand-side conditions should be clearer, currently they're really short & probably need a little more text in them.

E.g.:
* "Executables" - shared libraries are still executables. Should we say "Executables (that aren't shared libraries)"
* "for x86_64" - we originally said "on most x86 architectures (amd64, i386 and x32)". Which should we say?

I'd rather the condition be crystal-clear, since it impacts what people use.

Other than that, this is great. I'm thinking we should merge this *now*, and then argue over tweaks to it.

@thomasnyman - thoughts?

thomasnyman:
I agree. We can discuss the exact wording for the conditions in the next call. I'll merge this iteration now.

PR #281, opened 2023-11-08
Add note on _FORTIFY_SOURCE interaction with malloc_usable_size

siddhesh:
Buggy applications that use malloc_usable_size to *use* the residual size in the block may crash with _FORTIFY_SOURCE.  Add suggestions to deal with them.


thomasnyman:
lgtm

PR #282, opened 2023-11-08
Modify rationale for separate debug info

david-a-wheeler:
The current text *way* oversells the value of separate debug info. Depending on separating debug info for security is a terrible idea. Decompilers and similar tools work quite well even when debug info is separate. Also, many of the readers are thinking about open source software, and clearly trying to hide the source code when it's public doesn't make any sense (!).

So I suggest emphasizing the *size* argument (which is uncontroversial), then briefly noting the other argument but emphasizing that hiding debug info doesn't really prevent decompilation.

If you want to hide secrets, omitting debug info is a *terrible* way to do it :-). Let's not claim otherwise.


thomasnyman (approved):

thomasnyman:
lgtm

PR #301, opened 2023-11-22
Add clarification between production & instrumented test code, fixes #298

### LINKED ISSUE #298: In C/C++ hardening guide discuss instrumented test code vs production code
There seems to be a confusion between instrumented test code and production code.

When running software in production, you should use production code.

However, when testing, it is sometimes more complex. You typically want to test the production code, so that what you test is what you ship. However, if you only test the production code, you may miss some important problems. So it is often helpful to also generate executables that are especially introduced to warn about problems that would not be obvious in the production code version. I think that is the basis of some of the odd comments that some options shouldn't be used in production because they will make production code more reliable. Of course you want production code to be reliable!

I suggest adding a short section to describe the distinctions between these 2 kinds of executives. that will set up explaining the different options for the different situations.

david-a-wheeler:
Once we define these terms, we can use them as differentiators, enabling us to finally handle a lot of options that have been held up (because different situations have different needs).

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -44,6 +44,8 @@ See the discussion below for background and for detailed discussion of each opti
 
 Developers should ensure that their programs compile and pass their automated tests with all these options, e.g., by setting these as the default options. We encourage developers to consider it a bug if the program cannot be compiled with these options. Those who build programs for production may choose to omit some options that hurt performance if the program only processes trusted data, but remember that it's not helpful to deploy programs that that are insecure and rapidly do the wrong thing. Existing programs may need to be modified over time to work with some of these options.
 
+*Production* code is code intended for use in the real world with real effects; it should be maximally reliable and performant. *Instrumented test* code is code that is instrumented to improve defect detection and debuggability, and as such, often crashes more and is slower. Test processes should use both instrumented test code and production code.
+
 **Why do we need compiler options hardening?**
 
 Sadly, attackers today attack the software we use every day. Many programming languages' compilers have options to detect potential vulnerabilities while compiling and/or insert runtime protections against potential attacks. These can be important in any language, but these options are *especially* important in C and C++.


docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -44,6 +44,8 @@ See the discussion below for background and for detailed discussion of each opti
 
 Developers should ensure that their programs compile and pass their automated tests with all these options, e.g., by setting these as the default options. We encourage developers to consider it a bug if the program cannot be compiled with these options. Those who build programs for production may choose to omit some options that hurt performance if the program only processes trusted data, but remember that it's not helpful to deploy programs that that are insecure and rapidly do the wrong thing. Existing programs may need to be modified over time to work with some of these options.
 
+*Production* code is code intended for use in the real world with real effects; it should be maximally reliable and performant. *Instrumented test* code is code that is instrumented to improve defect detection and debuggability, and as such, often crashes more and is slower. Test processes should use both instrumented test code and production code.
```
thomasnyman, line 47: Should the new paragraph go before the previous paragraph since that already refers to "*production code*"?

The last sentence and in the new paragraph and the first sentence in the previous paragraph also seem to partly overlap. Maybe these could be combined for clarity:

> In this guide, we refer to *production code* for code intended for use in the real world with real effects; it should be maximally reliable and performant. We refer to *instrumented test code* for code that is instrumented to improve defect detection and debuggability, and as such, often crashes more and is slower. 
>
> Developers should ensure that both their production code and their instrumented test code pass their automated test with all these options. We encourage developers to consider it a bug if the program cannot be compiled with these options. Those who build production code may choose to omit some options that hurt performance if the program only processes trusted data, but remember that it's not helpful to deploy programs that that are insecure and rapidly do the wrong thing. Existing programs may need to be modified over time to work with some of these options.

Thoughts on this?

david-a-wheeler: Good idea. I'll tweak that text a little & create a new PR.

david-a-wheeler:
@thomasnyman - should we slip this clarification in now? I believe this is important for future changes.

gkunz (approved):
I agree that this should go in now.

thomasnyman (approved):

thomasnyman:
I agree. This looks good to me now.

PR #297, opened 2023-11-21
Add GCC and Clang version to -Wimplicit-fallthrough

thomasnyman:

PR #310, opened 2023-11-26
Add GCC and Clang version to -Wimplicit-fallthrough 

thomasnyman:
Corresponding change as in #297 for the recommended options table


thomasnyman:
Minor change, merging.

Issue #283, opened 2023-11-08
Add -Wbidi-chars=any to counter Trojan Source attacks


david-a-wheeler:
One kind of attack is "underhanded code" - this is maliciously misleading source code. Compilers can take steps to counter some of those. For a general discussion, see <https://www.ida.org/-/media/feature/publications/i/in/initial-analysis-of-underhanded-source-code/d-13166.ashx>

"Trojan Source" is a specific kind of underhanded code. This abuses Unicode right-to-left and left-to-right codes so that humans see one thing, but compilers interpret it differently. See https://news.ycombinator.com/item?id=29062982 and https://arxiv.org/pdf/2111.00169 and https://en.wikipedia.org/wiki/Trojan_Source . Trojan Source has CVE identifier(s) CVE-2021-42574 and CVE-2021-42694. Typically human review is the strongest defense against malicious code, and Trojan Source subverts this defense.

GCC has a `-Wbidi-chars` option that can counter this, see: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
By default it's "unpaired". A stronger value is `-Wbidi-chars=any`, though of course that means that source code can't include direction changes. Since such characters are rare, it seems reasonable to add that as a default. This would increase the likelihood that human review will interpret the source code the same way a compiler would.

I expect this to be a little more controversial, because this can make it more challenging to insert code comments in right-to-left languages like Hebrew and Arabic. In those rare cases you could override the option with a program-specific option. However, in the *vast* majority of cases, the only reason these markers are used in source code is an attack vector. Note that this option does *NOT* interfere with creating programs that can be *used* in right-to-left languages; best practice is to put such strings in separate file(s) and use an internationalization (i18n) framework like gettext to retrieve them.

Thoughts?

siddhesh:
Also, objdump has a `--unicode` option to escape unicode control characters, which was developed as part of the mitigation for CVE-2023-42574.  That was to look for symbol/section names with unicode control characters.

[...]

PR #317, opened 2023-11-29
Add threat model, goals, objectives

david-a-wheeler:
These are hinted at, but they should be clearly stated. This will make it much easier to justify the options below (including future options).

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -103,7 +103,7 @@ We are not trying to counter software whose source code is clearly written to be
 Given these goals, this guidance has the following objectives:
 
 1. *Minimize* the likelihood and/or impact of vulnerabilities that are released in production code.
-2. *Maximize* the detection and/of vulnerabilities during compilation or test (especially when using instrumented test code), so 
+2. *Maximize* the detection of vulnerabilities during compilation or test (especially when using instrumented test code), so they can be repaired before release.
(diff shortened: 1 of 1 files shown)


thomasnyman:
I think this should be introduced together with changes that address #283. I would also expect mitigating memory-safety issues to be mentioned in the threat model. I would argue that is a more obvious use of the guide, but I don't have any objections for addressing underhanded code too in the threat model.

david-a-wheeler:
**[thomasnyman](https://github.com/thomasnyman)**:

> I think this should be introduced together with changes that address https://github.com/ossf/wg-best-practices-os-developers/issues/283.

With respect, I think this needs to be added *first*. First, we need to make sure people understand the goals - otherwise it's not obvious *why* changes support the goals.

> I would also expect mitigating memory-safety issues to be mentioned in the threat model. I would argue that is a more obvious use of the guide, but I don't have any objections for addressing underhanded code too in the threat model.

That's a good idea. I think I can add a sentence at the end of a paragraph to make that clear *without* adding a lot of text.

david-a-wheeler:
[thomasnyman](https://github.com/thomasnyman) - I've added a sentence to expressly note the importance of memory safety. I made a few minor tweaks in this section since I was there anyway.

david-a-wheeler:
Note: I view this PR as a blocker for https://github.com/ossf/wg-best-practices-os-developers/issues/283 . Until we agree on the threat model we're supporting, it's hard to determine whether or not some specific flags are relevant.

thomasnyman:
Turned inline links into references in footnotes to be consistent with the established style.

thomasnyman:
Fixed broken link to Wikipedia article introduced in 6dff9b8abe1132648203a92170bb259fb884c2f9

thomasnyman (approved):

Issue #267, opened 2023-10-19
Add -mshstk to the C/C++ Compiler Hardening Guide?


david-a-wheeler:
I think we should consider adding `-mshstk` to the C/C++ Compiling Hardening guide.

It is recommended in [Use compiler flags for stack protection in GCC and Clang](https://developers.redhat.com/articles/2022/06/02/use-compiler-flags-stack-protection-gcc-and-clang#safestack_and_shadow_stack), where it notes: "This form of protection splits the stack into two distinct areas, storing precious variables and user variables in non-contiguous memory areas. The goal is to make it more difficult to smash one of the stacks from the other. This process occurs in hardware (e.g., x86_64 shadow stack) or in software such as LLVM's SafeStack."

This `-mshstk` flag is supported in clang and GCC. [GCC 9.1.0 documents](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/x86-Options.html) it as follows: "The -mshstk option enables shadow stack built-in functions from x86 Control-flow Enforcement Technology (CET)."

I'm not recommending adding Clang's SafeStack, which it adds through this flag: `-fsanitize=safe-stack`. [This StackOverflow answer](https://stackoverflow.com/questions/39371514/stack-protect-equivalent-in-clang-compiler) suggests there are problems with that with shared libraries. That may be incorrect (please let me know!), but until we're confident that problems are rare, I'm not confident in recommending it.

thesamesam:
The distinction between `-fcf-protection` and `-mshstk` is poorly covered in compiler documentation which doesn't help.

For our Hardened profiles in Gentoo, we pass `-fcf-protection` at least.

thomasnyman:
The built-in functions are, in my opinion, only needed if you are building a shadow stack support into C standard libraries or other code which may be doing unconventional management of the stack. 

I would suggest we cover `-mshstk` in the _Considered Compiler Options_ table (for completeness) but recommend `-fcf-protection`.

Issue #245, opened 2023-10-05
Compiler Hardening Guide should include -ftrivial-auto-var-init


thomasnyman:
Brought up during the [C/C++ Compiler BP Call 2023-09.27](https://docs.google.com/document/d/1UClGUnOSkOH_wab6Lx43KUdkaK37L8sbWJ_GPZvc1YY/edit#bookmark=id.863n4hp4nw9q).

The `-ftrivial-auto-var-init=[uninitialized|pattern|zero]` option will make the compiler initialize automatic variables with either a pattern or with zeroes (depending on the chosen option argument) to increase program security by preventing uninitialized memory disclosure and use.

This flag is supported since [GCC 12](https://gcc.gnu.org/gcc-12/changes.html) and [Clang 8.0](https://releases.llvm.org/8.0.0/tools/clang/docs/ReleaseNotes.html#major-new-features).

## Considerations and references
- [Patch and thread on GCC mailing list](https://gcc.gnu.org/pipermail/gcc-patches/2021-February/565514.html)
- [GCC Documentation](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Optimize-Options.html#index-ftrivial-auto-var-init)
- [Thread in LLVM Phabricator](https://reviews.llvm.org/D54604?id=174471)
- [Potential performance regressions caused by `-ftrivial-auto-var-init`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111523)
- [Endorsement from the Linux kernel community](https://lore.kernel.org/lkml/CAHk-=wgTM+cN7zyUZacGQDv3DuuoA4LORNPWgb1Y_Z1p4iedNQ@mail.gmail.com/)

siddhesh:
See also: performance regression in systemd due to this option: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111523

thesamesam:
There's nuance as to whether or not one should use `zero` or `pattern` as well:
* https://github.com/llvm/llvm-project/issues/44842
* https://inbox.sourceware.org/gcc-patches/8A3E5AA3-0785-4C2E-B75B-9388B703FFEA@gmail.com/T/#ma451313793180e6699a03d0f92ff11a1e08dd1b4

`zero` benefits from dead store elimination as well as not "looking valid" so it's possibly more likely it'll take some innocuous path rather than one which tries to operate on real data and write to disk or something.

[...]

PR #323, opened 2023-11-30
Add `-fno-delete-null-pointer-checks` and friends

david-a-wheeler:
Add the following:

* `-fno-delete-null-pointer-checks`
* `-fno-strict-overflow`
* `-fno-strict-aliasing`
* `-ftrivial-auto-var-init`

This is a redo/re-implementation of this pull request, per various comments:
https://github.com/ossf/wg-best-practices-os-developers/pull/139

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -173,9 +173,9 @@ Table 2: Recommended compiler options that enable run-time protection mechanisms
 | [`-fPIE -pie`](#-fPIE_-pie)                                                               |   Binutils 2.16<br/>Clang 5.0.0    | Build as position-independent executable. Can impact performance on 32-bit architectures.                                                   |
 | [`-fPIC -shared`](#-fPIC_-shared)                                                         | < Binutils 2.6<br/>Clang 5.0.0[^Guelton20] | Build as position-independent code. Can impact performance on 32-bit architectures.                                                         |
 | [`-fno-delete-null-pointer-checks`](#-fno-delete-null-pointer-checks)                     | GCC 3.0<br/>Clang 7.0.0            | Force retention of null pointer checks                                                       |
-| [`-no-strict-overflow`](#-fno-strict-overflow)                                            |                                    | Integer overflow may occur                                                                   |
-| [`-no-strict-aliasing`](#-fno-strict-aliasing)                                            |                                    | Do not assume strict aliasing                                                                |
-| [`-trivial-auto-var-init`](#trivial-auto-var-init)                                            |                                | Perform trivial auto variable initialization                                                 |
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Note: There are a number of unrelated markdownlint errors.  We should fix them, but I recommend fixing them in a *separate* branch since they are unrelated to this proposal.

david-a-wheeler:
@[thomasnyman](https://github.com/thomasnyman) @[gkunz](https://github.com/gkunz) - here are several new flags that start using the "production code vs. instrumented test code" distinction. Now that we have a path forward, let's add some flags that use it.

david-a-wheeler:
I don't know when these flags were added in GCC and Clang. I know that others are really good at finding out!

david-a-wheeler:
This also fixes: https://github.com/ossf/wg-best-practices-os-developers/issues/267

I think the problem earlier was we were trying to use the same options for *both* production code *and* for instrumented test code. Once you split things up, I think it's clearer.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -37,6 +37,8 @@ When compiling code in any of the situations in the below table, add the corresp
 | for shared libraries | `-fPIC -shared`                |
 | for x86_64           | `-fcf-protection=full`         |
 | for aarch64          | `-mbranch-protection=standard` |
+| for instrumented test code | `-ftrivial-auto-var-init=pattern` |
```
thomasnyman, line 40: As`-ftrivial-auto-var-init` interferes with unitialized memory tracking in tools such as Valgrind and AddressSanitizer I don't. think it should be enabled for instrumented test code. There was a discussion on this in [this comment to #245](https://github.com/ossf/wg-best-practices-os-developers/issues/245#issuecomment-1785479519).

thomasnyman: @thesamesam Thanks, I was thinking of MSAN too, not ASAN. The `-ftrivial-auto-var-init` does interfere with MSAN as can be seen here: https://godbolt.org/z/6qTPz9n6h

david-a-wheeler: If you don't use MSAN, is this still best to not use? I think many developers don't use MSAN.

david-a-wheeler: It'd be easy to remove the `=pattern` case. That's probably the best strategy, at least for now.

david-a-wheeler: Ok, let's do that. We can discuss =pattern in the details, without recommending its use in particular.

david-a-wheeler:
I believe everything is ready to go in this pull request.

I propose merging this. Is there agreement?

PR #324, opened 2023-11-30
Mend fences - use 3 not 4 characters

david-a-wheeler:
Per the CommonMark spec, fenced code blocks should use *3* characters not *4*. Some processors will quietly allow the other, but it can lead to problems, so let's use the standard format.

For more information, see:
<https://spec.commonmark.org/0.30/#fenced-code-blocks>.


thomasnyman (approved):
lgtm

thomasnyman:
I made to corresponding change to regular expression in the Makefile in 786b4a87bb0ee34d38bfce2dc1b745192643e5d9 to avoid breaking it with the corresponding change in the guide (as noted in @gkunz's review for [#325](https://github.com/ossf/wg-best-practices-os-developers/pull/325#discussion_r1417308687)).

PR #325, opened 2023-12-01
Improve Makefile so we can more flexibly test options

david-a-wheeler:


thomasnyman (approved):
lgtm

docs/Compiler-Hardening-Guides/Makefile
```diff
@@ -12,8 +12,10 @@
 # Recent POSIX adds "!=" and "::=" but they are not yet universally supported,
 # e.g., Apple MacOS only supplies an obsolete version of GNU Make.
 
+ifndef CFLAGS_HARDENING
 CFLAGS_HARDENING := $(shell sed -e '1,/~~~~/d' -e '/~~~~/,$$d' -e 's/\\$$//' \
```
gkunz, line 16: There is a dependency to #324: We need to modify the number of tilde characters here, too:

```suggestion
CFLAGS_HARDENING := $(shell sed -e '1,/~~~/d' -e '/~~~/,$$d' -e 's/\\$$//' \
```

thomasnyman: Fixed in 786b4a87bb0ee34d38bfce2dc1b745192643e5d9 merged with #324.

PR #329, opened 2023-12-01
best -> good

david-a-wheeler:
I don't think we can easily argue that this particular configuration is the "best balance" in all cases. As Disconnect3d from Trail of Bits noted in Slack, "citation needed :-)". Trade-offs are *way* too complicated to baldly state something is always the "best balance".

So instead, just say it's a "good balance". There may be a better balance for your circumstance, but since this is a good balance, we can recommend it.


disconnect3d (approved):
Makes sense to me :).

Fwiw, there are some performance analyses made in 2017 by [Zatoichi](https://zatoichi-engineer.github.io/blog), here is one for stack smashing protection, but they didn't take into account the ssp buffer size param: https://zatoichi-engineer.github.io/2017/10/04/stack-smashing-protection.html

(I will share this link as well on the OpenSSF #wg_best_pratices_compilers slack channel, and probably make another PR or issue regarding adding some reference links)

thomasnyman (approved):
lgtm

PR #345, opened 2023-12-27
More clearly distinguish -Werror with and without selectors

david-a-wheeler:
It's *fine* to use -Werror with selectors that shouldn't ever occur. The problem is -Werror without selectors.
Modify the text to clearly distinguish these two cases.


thomasnyman:
Made the additions consistently use 'blanket form' and 'selective form' when referring to the `-Werror` variants and streamlined the text in the synopsis by reordering the existing paragraphs to avoid some repetition.

PR #357, opened 2024-01-22
Add options to require use of Modern C (fixes #339)

### LINKED ISSUE #339: Require use of Modern C
It's possible to use flags to enforce use if "modern" C (C89 or later). More info here:

https://fedoraproject.org/wiki/Changes/PortingToModernC

Perhaps those should be added. Some changes may cause autoconf problems, we shoukd document how to address them, e.g., define exit().

david-a-wheeler:
This adds some `-Werror` flags to help people switch to Modern C instead of some problematic legacy constructs.

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -38,8 +38,9 @@ When compiling code in any of the situations in the below table, add the corresp
 | for x86_64           | `-fcf-protection=full`         |
 | for aarch64          | `-mbranch-protection=standard` |
 | for production code  | `-fno-delete-null-pointer-checks -fno-strict-overflow -fno-strict-aliasing -ftrivial-auto-var-init=zero` |
+| for disabling obsolete C constructs | `-Werror=implicit -Werror=incompatible-pointer-types -Werror=int-conversion` |
 
-We recommend developers to additionally use a blanket [`-Werror`](#-Werror) to treat all warnings as errors during development. However, `-Werror` should not be used in this blanket form when distributing source code, as this use of `-Werror` creates a dependency on specific toolchain vendors and versions. The selective form[`-Werror=`*`<warning-flag>`*](#-Werror-flag) that promotes specific warnings as error in cases that should never occur in the code can be used both during development and when distributing sources.
+We recommend developers to additionally use a blanket [`-Werror`](#-Werror) to treat all warnings as errors during development. However, `-Werror` should not be used in this blanket form when distributing source code, as this use of `-Werror` creates a dependency on specific toolchain vendors and versions. The selective form[`-Werror=`*`<warning-flag>`*](#-Werror-flag) that promotes specific warnings as error in cases that should never occur in the code can be used both during development and when distributing sources. For example, we encourage developers to promote warnings regarding obsolete C constructs removed by the 1999 C standard to errors (see the "for disabling obsolete C constructs" in the above table). These options often cannot be added by those who independently build the software, because the options may require non-trivial changes to the source code.
 
 See the discussion below for [background](#background) and for [detailed discussion of each option](#recommended-compiler-options).
 
@@ -185,6 +186,7 @@ Table 1: Recommended compiler options that enable strictly compile-time checks.
 | [`-Wtrampolines`](#-Wtrampolines)                                             |         GCC 4.3          | Enable warnings about trampolines that require executable stacks                    |
 | [`-Wimplicit-fallthrough`](#-Wimplicit-fallthrough)                           |         GCC 7<br>Clang 4.0   | Warn when a switch case falls through                                           |
 | [`-Werror`](#-Werror)<br/>[`-Werror=`*`<warning-flag>`*](#-Werror-flag)       | GCC 2.95.3<br/>Clang 2.6 | Treat all or selected compiler warnings as errors. Use the blanket form `-Werror` only during development, not in source distribution. |
+| [`-Werror=implicit`](#-Werror=implicit)<br/>[`-Werror=incompatible-pointer-types`](#-Werror=incompatible-pointer-types)<br/>[`-Werror=int-conversion`](#-Werror=int-conversion)<br/> | GCC 2.95.3<br/>Clang 2.6 | Treat obsolete C constructs as errors |
 
 Table 2: Recommended compiler options that enable run-time protection mechanisms.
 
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@fweimer - thoughts?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -38,6 +38,9 @@ When compiling code in any of the situations in the below table, add the corresp
 | for x86_64           | `-fcf-protection=full`         |
 | for aarch64          | `-mbranch-protection=standard` |
 | for production code  | `-fno-delete-null-pointer-checks -fno-strict-overflow -fno-strict-aliasing -ftrivial-auto-var-init=zero` |
+| for project developers | `-std=gnu2x -Werror=implicit-function-declaration -Werror=int-conversion -Werror=incompatible-pointer-types -Werror=old-style-definition -Werror=implicit-int` |
```
thomasnyman, line 41: Recommendations on language dialects go beyond how I understand the scope of the guide. IMHO that is better left up to individual projects, which may have other considerations, such as portability, that go beyond security aspects.

`gnu2x` in particular is not a good recommendation as [that name has been deprecated in favor `gnu23` in GCC](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html) (and would not work for C++ code either).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -38,6 +38,9 @@ When compiling code in any of the situations in the below table, add the corresp
...
+
+The options marked for project developers are options that should be added by the developers of the relevant project. These options often cannot be added by those who independently build the software, because they may require non-trivial changes to the source code.
```
thomasnyman, line 43: Isn't this recommendation contradicting the next paragraph regarding the use of blanket `-Werror` for development? I think we would need to prioritize one over the other to be consistent here. Maybe the recommendations for specific warnings to promote to errors could go into the [Synopsis for `-Werror`](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++#synopsis-5). We already mention `-Werror=trampolines` and `-Werror=conversion` there.

david-a-wheeler:
Per discussion:

> for project developers -> for disabling obsolete C constructs by project developers

Also, note within `-std=` that we add this option to enable disabling obsolete C constructs.

david-a-wheeler:
Hmm, `for disabling obsolete C constructs by project developers` is really long in the table. How about just `for disabling obsolete C constructs` and then explaining further in text?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Table entry rename resolves this I think.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: The reason for `gnu2x` is documented in the proposed text, it's more portable.

HOWEVER: I incorrectly remembered *why* `-std` was being proposed by Weimer. I think this argument makes sense, but it should be explained. I've added that rationale to this proposed text.

david-a-wheeler:
Fair point, we can drop `-std=gnu2x`.

> I'm not sure if it's a good use of programmer time to change void foo(); into void foo(void);

The broader problem is that in versions of C up to now (though not C++), `foo()` essentially disables all type-checking of parameters *and* causes conversions that surprise many people (e.g., float is silently cast to double).

That said, I can definitely see the "too hard" argument, as well as the "experimental argument".

thomasnyman:
Reworked the proposed text in b835a1bc05dcaa19993f62d2f1ca38225b5d3fcf as per the discussion here.

Summary of changes:
- Remove `-std=gnu2x` and `-Werror=old-style-definition`
- Consolidate `-Werror=implicit-function-declaration` and `-Werror=implicit-int` int `-Werror=implicit` supported by both GCC and Clang
- Move text in tl;dr as part of `-Werror` description to ensure the order the options are introduced in the text is consistent
- Move description of options from the `-Werror` description into a section of their own with a summary of the rationale in https://fedoraproject.org/wiki/Changes/PortingToModernC

thomasnyman:
Addressed @siddhesh's feedback from the [C/C++ Compiler BP Guide Call on 2024-02-14](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.cgn2l879w61l) that the text should refer to the Changes in GCC 14 and Clang 15 by compiler version. Also did minor fixes to footnote formatting.

thomasnyman:
Squashed the commits in this PR. Will propose to merge this in the [C/C++ Compiler BP Guide Call on 2024-02-28](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.pqc49td4xl9p).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -342,16 +344,56 @@ The C17 standard[^C2017] does not provide a mechanism to mark intentional fallth
...
+| <span id="-Werror=int-conversion">`-Werror=int-conversion`</span>                         | GCC 2.95.3<br/> Clang 2.6  | Treat implicit integer to pointer and pointer to integer conversions as errors                   |
+
+#### Synopsis
+
+Make the compiler treat obsolete C constructs as errors.
+
+The ISO/IEC 9899:1999 standard, commonly referred to as C99, removed several backwards compatibility features, such as implicit function declarations, implicit return types and certain implicit type conversion from the C language. Such implicit declarations[^DCL31-C] and type conversions[^INT36-C] can be considered dangerous for the correctness and security of C code as they lead to less stringent type checking and consequently can introduce unexpected and erroneous behavior. C99 and later dialects of C consider these issues as errors by default. However, modern compilers still accepts these obsolete construct by default unless instructed to pedantically give errors whenever the base standard requires them.
```
david-a-wheeler, line 369: s/still accepts/still accept/

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -342,16 +344,56 @@ The C17 standard[^C2017] does not provide a mechanism to mark intentional fallth
...
+| <span id="-Werror=int-conversion">`-Werror=int-conversion`</span>                         | GCC 2.95.3<br/> Clang 2.6  | Treat implicit integer to pointer and pointer to integer conversions as errors                   |
+
+#### Synopsis
+
+Make the compiler treat obsolete C constructs as errors.
+
+The ISO/IEC 9899:1999 standard, commonly referred to as C99, removed several backwards compatibility features, such as implicit function declarations, implicit return types and certain implicit type conversion from the C language. Such implicit declarations[^DCL31-C] and type conversions[^INT36-C] can be considered dangerous for the correctness and security of C code as they lead to less stringent type checking and consequently can introduce unexpected and erroneous behavior. C99 and later dialects of C consider these issues as errors by default. However, modern compilers still accept these obsolete construct by default unless instructed to pedantically give errors whenever the base standard requires them.
```
thesamesam, line 369: Implicit function declarations mean you get no fortification via `_FORTIFY_SOURCE` (which I mention at https://wiki.gentoo.org/wiki/Modern_C_porting#Is_this_cosmetic.3F).

david-a-wheeler:
[thesamesam](https://github.com/thesamesam) [33 minutes ago](https://github.com/ossf/wg-best-practices-os-developers/pull/357#discussion_r1594512075) - we've merged this pull request already. I suggest creating a new pull request instead.

thesamesam:
I was hoping to not have to make a PR for it, but alright, I will do.

thesamesam:
Done as https://github.com/ossf/wg-best-practices-os-developers/pull/357.

PR #363, opened 2024-02-14
Move improved compiling compilers section to end of document

### LINKED ISSUE #335: Should there be a "what should you do when compiling compilers" section at all?
Users don’t compile their compiler, generally speaking. So having this section is a bit surprising. But also, the section is pretty sparse in terms of its advice. Perhaps it makes sense to remove the section rather than expand it? Folks packaging applications for distros are basically the only target audience I can think of for this section, and they’ve already got their own best practices that will depend on the distro, I suspect.

thomasnyman:
Fixes #335


thomasnyman:
Will propose to merge this in the [C/C++ Compiler BP Guide Call on 2024-02-28](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.pqc49td4xl9p).

PR #370, opened 2024-03-08
Create initial stub for lab exercises

david-a-wheeler:


david-a-wheeler:
This is work in progress; do not merge yet.

wesleytodd:
Hey @david-a-wheeler! Happy to help review or provide feedback on this effort. I am not sure I understand the goals of this work in the OSSF, but also if there are good well maintained examples it would be good to cross link. Would there be any benefit to having the Express project actually "own" examples like this which align with your security recommendations?

david-a-wheeler:
I don't know if it makes sense for Express to "own" them or not, but I'm always delighted to discuss collaboration opportunities. Ensuring that example is secure code is, of course, also a plus :-). If nothing else, eyes to ensure that our secure code examples are actually secure & work properly would be GREAT.

One complication is that I'm trying to write labs assuming that the developer doesn't know JavaScript specifically - I just assume they can read *some* kind of language with an Algol-inspired syntax. That may not be a reasonable goal.

david-a-wheeler:
NOTE: HOWEVER: I intend to merge this lab & supporting infrastructure into the WG repo. It will be in the /docs/labs/ area.

That does *NOT* mean that this working group approved them!!

However, I need to be able to experiment & test to see what happens when these get published (and fix the inevitable problems). It's the best & easiest way to test it out, & I don't think it'll be a problem since "normal" users won't even see these them (they'd have to know the "secret" URLs). If someone has an issue with this, please let me know ASAP.

david-a-wheeler:
Another complication: Express is really sparse, and doesn't include its own validators, so I had to pick one. For simplicity, I chose express-validator. But that means these examples require that specific express-validator library as well as Express itself.

david-a-wheeler:
I've added a disclaimer to make it clear that these aren't approved by the education SIG nor the Best Practices WG yet. Making things clear seems like a good idea :-).

david-a-wheeler:
I see no objections here or on Slack, I added a disclaimer, and we can always undo this. So I'm merging this.

david-a-wheeler:
@wesleytodd (and others):

What would be great would be review of this draft lab at <https://best.openssf.org/labs/input1.html>. Is the code, once fixed up, a decent example of input validation?

The idea is that the user would read the instructions and work out how to fill in the missing material (in this case, something like `query('id').isInt({min: 1, max: 9999}),` ). I'm looking for simple examples to start with. I'm using integers here because validation of integers is simpler.

wesleytodd:
> I don't know if it makes sense for Express to "own" them or not, but I'm always delighted to discuss collaboration opportunities.

Same! I think when I wrote that I did not have enough context on your goals. Now I do

> One complication is that I'm trying to write labs assuming that the developer doesn't know JavaScript specifically....That may not be a reasonable goal.

This is for sure something we assume with most docs on Express related projects. I am not sure how valuable that level of generic docs are for most people, but also it might be that our target audiences are a bit different.

> Another complication: Express is really sparse, and doesn't include its own validators, so I had to pick one. For simplicity, I chose express-validator. But that means these examples require that specific express-validator library as well as Express itself.

Express does most (there is some history here which makes it not "clean") things via middleware. Many of which are "third party" like `express-validator`. We (the Express project or OpenJS Foundation) don't have anything to do with that package specifically. That doesn't mean anything really, especially doesn't indicate anything about the quality of the work, but we do plan on working toward some more of these "key features" having some level of "officially supported" flavor. In this case, ]I wrote an OpenAPI middleware](https://github.com/wesleytodd/express-openapi) which uses `ajv` under the hood which provides custom validators and is a very well maintained project. I plan to move this into the official Express org once we get some logistics sorted out.

I have no idea if this is valuable context for your work, but figured it is worth sharing just in case. 

> What would be great would be review of this draft lab at https://best.openssf.org/labs/input1.html. Is the code, once fixed up, a decent example of input validation?

I can take a look, but I personally think that when trying teach "generic best practice" it is a bad idea to add abstractions. I would personally much rather see simple JS than rely on learners then going to read some abstraction's docs to understand what it is really doing.

For this example, I did the leg work of tracking down what `express-validator` is [doing under the hood](https://github.com/express-validator/express-validator/blob/master/src/chain/validators-impl.ts#L247) and it is actually [implemented in a transitive](https://github.com/validatorjs/validator.js/blob/4197b8632522818164cc05d8d6cb44c98eb3decd/src/lib/isInt.js#L6). Frankly the fact that it is using a regex to validate it is an integer is almost certainly not what I would expect and also if a learner tried to track this down I think it would be hard to find and also likely *very* confusing (if not sending them in entirely the wrong direction 😭).

Take that opinion purely as my own and do with it what you will (ignoring it is perfectly alright 😄).

david-a-wheeler:
@wesleytodd :

First, I *really* appreciate your thoughts. Thank you!

> > What would be great would be review of this draft lab at https://best.openssf.org/labs/input1.html. Is the code, once fixed up, a decent example of input validation?

> I can take a look, but I personally think that when trying teach "generic best practice" it is a bad idea to add abstractions. I would personally much rather see simple JS than rely on learners then going to read some abstraction's docs to understand what it is really doing.

I agree we want to *minimize* abstractions, but I also want to make this somewhat realistic. If we wanted to eliminate all abstractions we'd write in machine code :-). Anyone writing server-side code *should* use a framework, not write from scratch, and a framework is an abstraction. Similarly, we're encouraging people to use common validators where they can, because validating some data types is harder than it looks (so it's better to reuse a library that's good at it). Given that, we probably ought to practice what we preach.

> For this example, I did the leg work of tracking down what express-validator is doing under the hood and it is actually implemented in a transitive. Frankly the fact that it is using a regex to validate it is an integer is almost certainly not what I would expect and also if a learner tried to track this down I think it would be hard to find and also likely very confusing (if not sending them in entirely the wrong direction 😭).

That's unsurprising to me, and it seems reasonable (as long as it's the *right* regex).

wesleytodd:
> I agree we want to minimize abstractions, but I also want to make this somewhat realistic. If we wanted to eliminate all abstractions we'd write in machine code

Totally, and I would never argue differently. The problem always finding the right one, and popularity is not the main criteria. So in this case I would rather recommend folks use JSON schema and a corresponding validator as a "best practice" than use a JS specific abstraction package. That way the same thing transcends the language tie in even, which I think helps with your use case as described above.

> That's unsurprising to me, and it seems reasonable (as long as it's the right regex).

haha, yeah that is certainly key. That said the language has specifications for this and it is not a regex 😆.

PR #438, opened 2024-03-27
Add -Wbidi-chars=any, fixes #283

### LINKED ISSUE #283: Add -Wbidi-chars=any to counter Trojan Source attacks

david-a-wheeler:


docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,52 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
 
 ---
 
+### Warn on Unicode Unicode bidirectional (bidi) override characters |
+
+| [`-Wbidi-chars=any`](#-Wbidi-chars) | GCC | Warn on Unicode bidirectional (bidi) override characters |
```
thomasnyman, line 401: This table row seems to have been misplaced

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,52 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
...
+
+| Compiler Flag                                                                              | Supported since            | Description                                                                                  |
+| ------------------------------------------------------------------------------------------ | ----------------------- | -------------------------------------------------------------------------------------------- |
+| <span id="--Wbidi-chars">`-Wbidi-chars=any`</span>                                | GCC | Fortify sources with compile- and run-time checks for unsafe libc usage and buffer overflows         |
```
thomasnyman, line 405: The GCC description of this option also covers variants that warn about bidi characters expressed via universal characters names? Would it make sense to recommend `-Wbidi-char=any,ucn` as that seems to be the most restrictive variant, and include the other variants, including the `unpaired` referred to from the synopsis in this table too? 

Some additional nits:
- The span id is incorrect 
- The correct description seems to be in the misplaced table row above 
- GCC version missing (`-Wbidi-chars` was added in [GCC 12](https://developers.redhat.com/articles/2022/04/25/new-c-features-gcc-12))

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,52 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
...
+
+<!-- Implemented in: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103026 -->
+
+[^Boucher2021]: [Boucher, Nicholas and Ross Anderson, 2021, "Trojan Source: Invisible Vulnerabilities"](https://arxiv.org/abs/2111.00169)
+
+[^gcc-Wbidi-chars]: GCC team, [Using the GNU Compiler Collection (GCC): Warning Options: `-Wbidi-chars`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wbidi-chars_003d),
+
+[^Wheeler2020]: [Wheeler, David A., April 2020, *Initial Analysis of Underhanded Source Code*](https://www.ida.org/-/media/feature/publications/i/in/initial-analysis-of-underhanded-source-code/d-13166.ashx).
+
+#### Performance implications
```
thomasnyman, line 429: So far we've omitted the performance implications sections for warnings since the guide states already in the [general description](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++#recommended-compiler-options) that the options in Table 1 don't incur performance penalties.

Maybe we can drop this one too for consistency and brewity.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,52 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
...
+
+None.
+
+#### When not to use?
+
+Do *not* use `-Wbidi-chars=any` in cases where some of the source code *is* expected to include bidirectional control characters. This is typically the case where some of the source code text is in Arabic or Hebrew. In such cases, use `-Wbidi-chars=unpaired` (the default) instead on GCC. This alternative still provides some defenses.
```
thomasnyman, line 435: Suggest rewording to 

> Arabic, Hebrew, or other right-to-left script.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,52 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
...
+
+"Trojan Source" is a specific kind of underhanded code that exploits the Bidirectional Algorithm in the Unicode Specification. Some natural languages (such as Arabic and Hebrew) are typically written right-to-left (RTL), while many others (such as English) are written left-to-right (LTR). Some texts must mix texts with different orders.  Unicode supports various control sequences to support this visual reordering. Unfortunately, attackers can use these control sequences so they can introduce vulnerabilities in source code that might not be visible to human reviewers.
+
+For more about Trojan Source see [trojansource.com](https://trojansource.codes/), which includes a copy of the paper "Trojan Source: Invisible Vulnerabilities" by Nicholas Boucher and Ross Anderson [^Boucher2021]. More context can be found [this Hacker News discussion](https://news.ycombinator.com/item?id=29062982) and [Wikipedia](https://en.wikipedia.org/wiki/Trojan_Source).  See, in particular, [CVE-2021-42574](https://nvd.nist.gov/vuln/detail/CVE-2021-42574) and [CVE-2021-42694](https://nvd.nist.gov/vuln/detail/CVE-2021-42684).
```
thomasnyman, line 413: Could this paragraph be turned into footnote and referred to also from the threat model? 

The references here would be relevant already there.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thomasnyman: Proposed changes addressing these suggestion in f62e99ec1b64bde0c92f873a167b9123e09cca66

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thomasnyman: Proposed changes to remove the performance implications section in 5620102c4a5378f39f2a305179fbba6567d7270f

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thomasnyman: Fixed in d8347c193b383ed360d1d6b32ec049e150a18518

david-a-wheeler: Maybe "Source text solely left-to-right & GCC" ?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -42,6 +42,7 @@ When compiling code in any of the situations in the below table, add the corresp
 | for aarch64          | `-mbranch-protection=standard` |
 | for production code  | `-fno-delete-null-pointer-checks -fno-strict-overflow -fno-strict-aliasing -ftrivial-auto-var-init=zero` |
 | for disabling obsolete C constructs | `-Werror=implicit -Werror=incompatible-pointer-types -Werror=int-conversion` |
+| No right-to-left text in source & using GCC | `-Wbidi-chars=any` |
```
thomasnyman, line 45: Start with "using GCC and only left-to-right text in source code"

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -394,6 +396,49 @@ Some tools, such as `autoconf`, automatically determine what the compiler suppor
 
 ---
 
+### Enable warnings for possibly misleading unicode bidirectional control characters
+
+| Compiler Flag                                                             | Supported since | Description                                                                                                                                    |
+| ------------------------------------------------------------------------- | ----------------| ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| <span id="-Wbidi-chars=any,ucn">`-Wbidi-chars=any,ucn`</span>             | GCC 12          | Enable warnings for any UTF-8 and UCN bidirectional control characters in comments, string literals, character constants, and identifiers       |
```
thomasnyman, line 403: Double-check whether the ucn option is relevant for security. 

Need test case for confirming whether the ucn option affects UCNs in general or just bidi UCNs.

thomasnyman: Based on [this test case](https://godbolt.org/z/McssM61Ye) `-Wbidi-char=any,ucn` only warns on UCNs that correspond to bidi characters in string literals, character constants, and identifiers. Non-bidi UCNs in comments, string literals, character constants, and identifiers are not affected, and neither are bidi UCNs in comments.

thomasnyman:
Updated PR as per extensive discussion in [C/C++ BP Guide Call 2024-03-28](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.35uvnbh21e8q).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Okay. There's a merge conflict. I'll fix that up.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: I reworded further. It turns out that Persian and Urdu are right-to-left. We'll mention both, in both lists of examples. That'll at least make it clear there are *multiple* writing systems relevant to this (and thus why "always forbidding" isn't the right approach).

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: Yes, done.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: It doesn't matter for *Trojan* Source, since they're clearly visible in the source code as `\uXXXX` or \U....`. See the patch: https://gcc.gnu.org/pipermail/gcc-patches/2021-November/583031.html

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
david-a-wheeler: They've been moved into the discussion about the Boucher paper.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -334,6 +336,47 @@ The C17 standard[^C2017] does not provide a mechanism to mark intentional fallth
...
+
+The GCC `-Wbidi-chars` option helps to counter Trojan Source attacks[^gcc-Wbidi-chars]. By default its value is `-Wbidi-char=unpaired`, which warns about improperly terminated bidi contexts (this should never happen in source code). However, this default is somewhat permissive.
+
+In many cases using `-Wbidi-char=unpaired` is a stronger defense. This option forbids *any* use of bidirectional control characters in comments, string literals, character constants, and identifiers, completely eliminating the Trojan Source attack. This setting is appropriate when bidi characters are *not* expected in the source code, and their only use would be as part of an attack on reviewers.
```
gkunz, line 356: fixing a copy-and-paste issue

```suggestion
In many cases using `-Wbidi-char=any` is a stronger defense. This option forbids *any* use of bidirectional control characters in comments, string literals, character constants, and identifiers, completely eliminating the Trojan Source attack. This setting is appropriate when bidi characters are *not* expected in the source code, and their only use would be as part of an attack on reviewers.
```

gkunz (changes requested):

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thomasnyman: Fixed in 8e7ee821391bed3fe7a0f9e80760012095a82fc0.

thomasnyman:
Squashed all commits into e76ac8d27d59646418781181b6a72e929f0e7858. Added @gkunz as co-author.

gkunz (approved):

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -334,6 +336,47 @@ The C17 standard[^C2017] does not provide a mechanism to mark intentional fallth
...
+Do *not* use `-Wbidi-chars=any` or `-Wbidi-chars=any,ucn` in cases where some of the source code *is* expected to include bidirectional control characters. This is typically the case where some of the source code text, e.g., comments, are in a right-to-left script such as Arabic, Hebrew, Persian, or Urdu. In such cases, use `-Wbidi-chars=unpaired` (the default) or `-Wbidi-chars=unpaired,ucn` instead.
+
+#### Additional Considerations
+
+It is best to use other static code analysis tools to also warn about Trojan Source, since it's not an issue developers typically consider. Some editors have mechanisms to warn about Trojan Source; using them is recommended where practical. However, it's sometimes difficult to verify whether developers and reviewers have used such tools.
+
+clang-tidy's `misc-misleading-bidirectional` check warns about unterminated bidirectional Unicode sequences, similar to GCC's `-Wbidi-char=unpaired`[^clang-tidy-bidi].
+
+[^clang-tidy-bidi]: LLVM team, [clang-tidy - misc-misleading-bidirectional](https://clang.llvm.org/extra/clang-tidy/checks/misc/misleading-bidirectional.html), Extra Clang Tools Documentation, 2024-03-28.
+
```
06kellyjac, line 377:
```suggestion
`clang-tidy` has multiple lints to help identify Trojan Source:

* `misc-misleading-bidirectional` check warns about unterminated bidirectional Unicode sequences, similar to GCC's `-Wbidi-char=unpaired`[^clang-tidy-bidi].
* `misc-confusable-identifiers` check warns about characters that are visually similar [^clang-tidy-confusable].
* `misc-misleading-identifier` check warns about bidirectional Unicode that can change the meaning of the code [^clang-tidy-misleading].

[^clang-tidy-bidi]: LLVM team, [clang-tidy - misc-misleading-bidirectional](https://clang.llvm.org/extra/clang-tidy/checks/misc/misleading-bidirectional.html), Extra Clang Tools Documentation, 2024-03-28.
[^clang-tidy-confusable]: LLVM team, [clang-tidy - misc-confusable-identifiers](https://clang.llvm.org/extra/clang-tidy/checks/misc/confusable-identifiers.html), Extra Clang Tools Documentation, 2024-03-28.
[^clang-tidy-misleading]: LLVM team, [clang-tidy - misc-misleading-identifier](https://clang.llvm.org/extra/clang-tidy/checks/misc/misleading-identifier.html), Extra Clang Tools Documentation, 2024-03-28.

```

I can change this to a table if necessary. Also IDK if clang-tidy has any more trojansource style lints.

I just tried the confusable example with `-Wbidi-char=any` and `-Wbidi-char=any,ucn` but I couldn't get it to warn. IDK if gcc has an equivalent or a disable all unicode

And for misleading I've also not been able to find a disable unicode completely for the code.

```c
#include <stdio.h>

short int א = (short int)0;
short int ג = (short int)12345;

int main() {
  int א = ג; // a local variable, set to zero?
  printf("ג is %d\n", ג);
  printf("א is %d\n", א);
}
```

Out:
```
ג is 12345
א is 12345
(also has some ltr rtl funniness in the raw output)
```

image

Issue #341, opened 2023-12-10
Vtable verification during compiling C++ code


Flo4152:
Vtable verification during compiling C++ code introduce verifying virtual function pointers at run time. This security feature is available on GCC with this option `-fvtable-verify` (https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtable-verify).

I don't know equivalent on clang compiler.

Ensuring the integrity of virtual function pointers prevents overwriting pointers on vulnerable binaries.

Do you think this option may have its place on "Compiler Hardening Guide"?

Regards,

david-a-wheeler:
This creates a small overhead on every method call that uses vtables. It is likely to be okay, but I'd like to know a little more about the attacks vs. the performance impact. Are there samples of the attacks & the performance impacts?

Flo4152:
I know about vtable Hijacking attack. This kind of attack is based on memory corruption such as heap overflow to overwrite pointer of a c++objects. These pointers are mostly C++ virtual function address (Class method). Controlling one of these pointers, attackers have full access to the flow of control of an application.

GCC offers option to add verification of vtable pointers at run time (-fvtable-verify) so the performance impacts exist. This performance is proportional to the object complexity of a C++ application : The more class methods, the greater the performance impacts.

Yes, I want to create a pull request on this topic. This is the first time I've done this, I need help. I think I should clone your repository and apply my modification?

Regards,

thomasnyman:
@Flo4152 You'll have the easiest time if you first [fork the ossf/wg-best-practices-os-developer repo](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo) to your own Github profile, apply your changes to your local fork, then [create a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork). 

You can clone the repository and edit it locally and push your changes to your fork, but Github also supports [editing files online] (https://docs.github.com/en/repositories/working-with-files/managing-files/editing-files).

david-a-wheeler:
The performance impact isn't just on startup, but on every C++ virtual call:

> "verifies at run time, for every virtual call, that the vtable pointer through which the call is made is valid for the type of the object, and has not been corrupted or overwritten. If an invalid vtable pointer is detected at run time, an error is reported and execution of the program is immediately halted."

[...]

PR #440, opened 2024-03-28
add GCC option for checking virtual table pointers

Flo4152:
GCC offers an option to add virtual table pointer verification at runtime (-fvtable-verify).

### DIFF
--- a/docs/labs/README.md
+++ b/docs/labs/README.md
@@ -21,6 +21,8 @@ Here are the labs available:
 * [csp1](csp1.html) - Content Security Policy (CSP)
 * [oob1](oob1.html) - Out-of-bounds (OOB)
 
+We also have a [template](template.html) available.
+
 ## Please contribute labs
 
 We would love to have people contribute relevant labs to help
@@ -36,6 +38,7 @@ See below for how to create labs and our lab roadmap.
 See [create labs](create_labs) if you want to learn how to create labs.
 In particular, that page will link to how to
 [create labs using checker](create_checker).
+We suggest using the [template](template.html) as a start.
 
 To submit new or updated labs, create a pull request on the
 [OpenSSF Best Practices WG](https://github.com/ossf/wg-best-practices-os-developers/)
--- a/docs/labs/create_checker.md
+++ b/docs/labs/create_checker.md
@@ -27,7 +27,9 @@ There are three basic tasks, which can be done by different people:
    [input1.html](input1.html), and modify it for your situation.
    See David A. Wheeler who can help you get started.
 
-The text below discusses these in more detail. You can also see our
+The text below discusses these in more detail.
+We suggest using the [template](template.html) as a start.
+You can also see our
 [potential future directions](#potential-future-directions).
 
 ## Creating the lab instructions and correct answer
@@ -45,6 +47,8 @@ do *not* assume they know any particular programming language.
 See [input1.html](input1.html),
 [input2.html](input2.html), and
 [csp1.html](csp1.html) for examples of how to do this.
+
+We suggest using the [template](template.html) as a starting point.
 
 ## Creating the lab HTML file
 
@@ -65,12 +69,9 @@ The system is implemented by the client-side JavaScript file `checker.js`.
 ### TL;DR
 
 An easy way implement a lab is to copy
-[input1.html](input1.html) and modify it for your situation.
+use our [template](template.html) and modify it for your situation.
 Modify the `expected0` section to have a sample expected answer, and
 `correct0` to have a full pattern for a correct answer.
-You can comment out the `info` section (put `#` in front of each line)
-to start.
-
 See [input1.html](input1.html) and [input2.html](input2.html)
 for examples.
 
--- a/docs/labs/csp1.html
+++ b/docs/labs/csp1.html
@@ -409,6 +409,11 @@ app.listen(3000);
 <button type="button" class="resetButton">Reset</button>
 <button type="button" class="giveUpButton">Give up</button>
 <br><br>
+<p>
+<i>This lab was developed by David A. Wheeler at
+<a href="https://www.linuxfoundation.org/"
+>The Linux Foundation</a>.</i>
+<br><br>
 <p id="correctStamp" class="small">
 <textarea id="debugData" class="displayNone" rows="20" cols="65" readonly>
 </textarea>
--- a/docs/labs/hello.html
+++ b/docs/labs/hello.html
@@ -174,6 +174,11 @@ For multi-line inputs, instead of <input id="attempt0" type="text" ...>, use
 <button type="button" class="resetButton">Reset</button>
 <button type="button" class="giveUpButton">Give up</button>
 <br><br>
+<p>
+<i>This lab was developed by David A. Wheeler at
+<a href="https://www.linuxfoundation.org/"
+>The Linux Foundation</a>.</i>
+<br><br>
 <p id="correctStamp" class="small">
 <textarea id="debugData" class="displayNone" rows="20" cols="65" readonly>
 </textarea>
--- a/docs/labs/input1.html
+++ b/docs/labs/input1.html
@@ -224,6 +224,11 @@ app.get('/invoices',
 <button type="button" class="hintButton">Hint</button>
 <button type="button" class="resetButton">Reset</button>
 <button type="button" class="giveUpButton">Give up</button>
+<br><br>
+<p>
+<i>This lab was developed by David A. Wheeler at
+<a href="https://www.linuxfoundation.org/"
+>The Linux Foundation</a>.</i>
 <br><br><!-- These go in the last form if there's more than one: -->
 <p id="correctStamp" class="small">
 <textarea id="debugData" class="displayNone" rows="20" cols="65" readonly>
--- a/docs/labs/input2.html
+++ b/docs/labs/input2.html
@@ -238,6 +238,11 @@ app.get('/parts',
 <button type="button" class="resetButton">Reset</button>
 <button type="button" class="giveUpButton">Give up</button>
 <br><br>
+<p>
+<i>This lab was developed by David A. Wheeler at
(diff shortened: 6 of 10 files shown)


thomasnyman:
Hi @Flo4152 and thank you for your contribution. The [OpenSSF Charter](https://openssf.org/about/charter/) requires all contributions to be accompanied by a [Developer Certificate of Origin](https://developercertificate.org/) (DCO) sign-off in the commit message that certifies the contribution is made according to the stipulations of the Linux Foundation DCO. Could you please read the DCO at the above link, and if you are able and willing to certify that the DCO requirements are met, amend the commit message with your sign-off.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -205,6 +205,7 @@ Table 2: Recommended compiler options that enable run-time protection mechanisms
 | [`-fno-strict-overflow`](#-fno-strict-overflow)                                           | GCC 4.2                            | Integer overflow may occur                                                                   |
 | [`-fno-strict-aliasing`](#-fno-strict-aliasing)                                           | GCC 2.95.3<br/>Clang 18.0.0        | Do not assume strict aliasing                                                                |
 | [`-ftrivial-auto-var-init`](#-ftrivial-auto-var-init)                                     | GCC 12<br/>Clang 8.0               | Perform trivial auto variable initialization                                                 |
+| [`-fvtable-verify=std`](#-fvtable-verify)                                     | GCC 4.9.4               | Enable run-time checks for virtual function pointers corruption (C++). Can impact performance.                                                  |
```
thomasnyman, line 208: Suggested rewording:

> Enable run-time type checks for C++ virtual function pointers. Can impact performance.

- Would move C++ to the description from the parenthesis to be consistent with the wording for [`-D_GLIBCXX_ASSERTIONS`](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++#-D_GLIBCXX_ASSERTIONS)
- The [GCC documentation specifes](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtable-verify) the check is validates the vtable pointer is correct for the type of object. It would be good to represent that in the short description.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,28 @@ https://godbolt.org/z/6qTPz9n6h
 
 ---
 
+### Vtable pointers check
```
thomasnyman, line 868: Use the first sentence of short description from Table 2 as title here to be consistent with the other option descriptions.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,28 @@ https://godbolt.org/z/6qTPz9n6h
...
+
+| Compiler Flag                                                       | Supported since     | Description                                  |
+|:--------------------------------------------------------------------|:-------------------:|:---------------------------------------------|
+| <span id="-fvtable-verify">`-fvtable-verify`</span> | GCC 4.9.4 | Enable run-time checks for virtual function pointers corruption (C++). Can impact performance. |
```
thomasnyman, line 872: We've generally tried to ensure that the options is depicted in the various tables it appears in is in a form where the option can be copy-pasted as is into a compiler command line. 

Since `-fvtable-verify` needs one of the `std`, `preinit` or `none` arguments, it would be better to include those variants in the correct calling form in this table, with the recommended argument first in the list. 

Also, rather than repeating the "Can impact performance" here, it would be better to have a separate "Performance implications" subsection with a more detailed description of what factors impact the performance degradation, as you note in the [associated issue](https://github.com/ossf/wg-best-practices-os-developers/issues/341#issuecomment-2017204937). (But shouldn't the performance impact scale with the number of virtual methods calls though, not the number of class methods?). 

Any references to benchmarks or documented cases performance degradation to get an idea of the scale here would be great.

Reading the option description in GCC implies that the `std` variant would already be the default. Is that the case? If so, then any performance impact of the the run-time check would already be part of the baseline performance.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,28 @@ https://godbolt.org/z/6qTPz9n6h
...
+
+#### Synopsis
+
+In object-oriented programming, object methods rely on dynamic dispatching to select the right virtual function to call at runtime. The vtable is the table of virtual function pointers linked to the code of the virtual function to be executed. The vtable can be overwritten by memory corruption. This option enables vtable checking at runtime for each virtual function calls. This check prevents virtual function pointers from being corrupted or overwritten, and stops the application if the pointer integrity check fails. This virtual table pointer check is based on the data structure built at application start-up.
```
thomasnyman, line 876: > The vtable can be overwritten by memory corruption.

This could name "vtable hijacking" explicitly as that is a well-established name for this particular type of attack.

> This option enables vtable checking at runtime for each virtual function calls.

Typo: "each virtual call" (singular)

> and stops the application if the pointer integrity check fails

Is this a SIGABRT as in the case for the stack protector functionality?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,28 @@ https://godbolt.org/z/6qTPz9n6h
...
+
+This option has three choices:
+
+- `std` - vtable checks based on data structure built after shared libraries have been loaded and initialized
+- `preinit` - vtable checks based on data structure built before shared libraries have been loaded and initialized
```
thomasnyman, line 881: What's the tradeoff between `std` and `preinit` and why would one choose one over the other?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,28 @@ https://godbolt.org/z/6qTPz9n6h
...
+- `none` - turns off the vtable checks.
+
+<!-- More information
+https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtable-verify
+-->
+
```
thomasnyman, line 887: When not to use?

thomasnyman (changes requested):

david-a-wheeler:
I'm very concerned about the performance impact:
https://github.com/ossf/wg-best-practices-os-developers/issues/341

GabrielDosReis:
> I'm worried if the performance impact is too much to be usable in production. Are there projects or Linux distributions or even a Linux distro package that uses this option (suggesting it's not too bad)? Also, I think we need to document the performance impact in this flag if it's option.

I share similar concerns.  

I understand it was used at some places, but it isn't free and the cost is not necessarily commensurate to what is sought.  Maybe, there are some realistic benchmarks to point to, to help programmers to assess when it is the right thing for them?

Flo4152:
@thomasnyman @david-a-wheeler @GabrielDosReis To answer to the question about performance penalty, in addition to Caroline Tice articles about performance penalties,  I think scripts and tools suites from Caroline Tice's commit offers ways to evaluate the number of vcall being verified.

The `-fvtv-counts` flags from GCC provides way to process source codes to count of vcalls being verified by the feature  `-fvtable-verify`. (https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtv-counts and https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libvtv/scripts/sum-vtv-counts.c;h=fc99498e70522a77645d27f01f4b3cdf754a8829;hb=2077db1be5b18b94a91095a3fb380bbc4a81e61b)

These counters offers a evaluation of performance penalty of vtable verification feature of GCC.

I'm working to add a description of  `-fvtv-counts` flags in my pull request.

Regards,

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -864,6 +865,60 @@ https://godbolt.org/z/6qTPz9n6h
...
+Additional debugging information is also available using GCC's `-fvtv-debug` flag. This flag provides information on binaries compiled with `-fvtable-verify` to gain a better understanding of the action during virtual table verification. It traces the virtual table pointers checked for each class of source code in a log file (vtv_set_ptr_data.log).
+
+The location of `-fvtv-counts` and `-fvtv-debug` log files is defined by the `VTV_LOGS_DIR` environment variable. The default location is the current directory.
+
+<!-- More information
+https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtable-verify
+https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=2077db1be5b18b94a91095a3fb380bbc4a81e61b
+https://gcc.gnu.org/wiki/cauldron2012?action=AttachFile&do=get&target=cmtice.pdf#page=38
+https://www.usenix.org/sites/default/files/conference/protected-files/sec14_slides_tice.pdf#page=14
```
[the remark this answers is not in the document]
Flo4152: The functionality requested on the IT forum leads to this ticket: https://sir.ext.ti.com/jira/browse/EXT_EP-10096

This request has been classified as "non-priority" since November 2020 and this issue is still open with no activity. I don't think this message is relevant as it hasn't been implemented yet.

thomasnyman (approved):

PR #447, opened 2024-04-02
Add sound effects for lab successes

david-a-wheeler:
Add some small sound effects when there is a lab success. The goal is to be (1) motivating, and (2) aid accessibility for those who cannot perceive the background color changes that we perform to indicate partial success.


ctcpip:
technically, the longer sound does not meet WCAG 1.4.2, but the issue is minor and probably fine

> aid accessibility for those who cannot perceive the background color changes that we perform to indicate partial success

per WCAG 1.4.1, can some other visual indicator be added for this?

david-a-wheeler:
I've gotten a lot of "please don't", so I'll quietly close this :-).

PR #449, opened 2024-04-02
Change lab html style to refer to best.openssf.org

david-a-wheeler:
Modify the lab HTML files to always try to load from best.openssf.org. This is no change for accesses to best.openssf.org, but it means that if you're using the lab locally *and* have internet access, you'll see the same thing. This is an improvement for local testing and individualized use.

Neither approach will allow the use of styles for disconnected operations. That said, the labs will still *work* they'll just look different.


david-a-wheeler:
Didn't work as expected. Closing.

david-a-wheeler:
Ah! I found the bug. Now it works.

PR #469, opened 2024-04-30
Update Compiler-Options-Hardening-Guide-for-C-and-C++.md

bagder:
make the fallthrough macro look like a function call

This makes it look like normal code and thus it keeps syntax checkers happy and automatic indentation still works etc.


david-a-wheeler:
Wait a second! I don't mind switching this to function call notation, but the sentence above it becomes wrong. The Linux kernel does *NOT* use function call notation. So the sentence above needs to say "inspired by" or "similar to".

david-a-wheeler:
We agreed at the meeting today to fix the text above it to reflect this change to function call notation. I created a fix, but accidentally merged it into main instead of its own branch. Since we agreed to do this today during our meeting, I think we just "let the quick fix ride". If anyone sees a problem with the quick fix, please create a PR to fix the fix :-).

Issue #326, opened 2023-12-01
Evaluate `-fsanitize-minimal-runtime` for possible hardening


david-a-wheeler:
In https://news.ycombinator.com/item?id=38478866 [nlewycky](https://news.ycombinator.com/user?id=nlewycky) reports the following:

> Also! While the usual sanitizer runtime libraries aren't security hardened for use in production environments, but for UBSan there's -fsanitize-minimal-runtime which switches to a different runtime library that is intended for this purpose (or use -fsanitize-trap=... instead, which executes an illegal instruction on error). Note that if your program terminates with a UBSan error, an attacker who can check whether your program terminated or not could use that as a primitive to leak data, so consider the security impact on your use case carefully. UBSan has a quite small performance impact when building with optimization, so you could deploy to production with it enabled, or parts of it enabled.

I'd like some feedback from the various distros & such. Is there a big drawback to this? It sounds potentially helpful, but I don't have any experience doing this for production.

SecurityCRob:
Has this been addressed by the C/C++ Compiler Hardening options guide?    @gkunz @thomasnyman @david-a-wheeler

thesamesam:
The equivalent in GCC is `-fsanitize-trap=undefined`. Per the [docs](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize-trap):
> The advantage of this is that the libubsan library is not needed and is not linked in, so this is usable even in freestanding environments.

PR #485, opened 2024-05-08
add GCC option for checking virtual table pointers

Flo4152:
add GCC option for checking virtual table pointers


Flo4152:
New pull request to add a section about virtual table checking section after synchronizing my repository with the base branch.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -211,6 +211,7 @@ Table 2: Recommended compiler options that enable run-time protection mechanisms
 | [`-fno-strict-aliasing`](#-fno-strict-aliasing)                                           | GCC 2.95.3<br/>Clang 18.0.0        | Do not assume strict aliasing                                                                |
 | [`-ftrivial-auto-var-init`](#-ftrivial-auto-var-init)                                     | GCC 12<br/>Clang 8.0               | Perform trivial auto variable initialization                                                 |
 | [`-fexceptions`](#-fexceptions)                                                           | GCC 2.95.3<br/>Clang 2.6           | Enable exception propagation to harden multi-threaded C code                                 |
+| [`-fvtable-verify=std`](#-fvtable-verify)                                                 | GCC 4.9.4                          | (C++) Enable run-time checks for virtual function pointers corruption. Can impact performance.|
```
thesamesam, line 214: IIRC, this option breaks ABI (at least if libstdc++ is built with it, which is needed for it to be more useful).

Also, I think GCC supports `-fsanitize=vptr` nowadays.

Flo4152:
@thomasnyman @david-a-wheeler This pull request is a duplication of #440. I close my first pull request by mistake during the synchronization of my repository.

Sorry about this mistake.

Regards,

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
Flo4152: This risk of ABI breakage is described in the Gentoo manual (USE flag - https://packages.gentoo.org/useflags/vtv). This warning seems to be related to maintaining full package compatibility and stability requirements throughout the lifetime of a system (library dependencies and link editing issues). This information provides a realistic argument for answering @david-a-wheeler's question about the use of `-fvtable-verify` in a real environment (https://github.com/ossf/wg-best-practices-os-developers/issues/341#issuecomment-2032613094). 

@thesamesam is right, the use of `-fvtable-verify` may not be recommended in many real-life situations. Adding this flag to the Guide may require, at a minimum, additional research and I can't provide any arguments to justify adding this flag to the Guide. I can only provide a few examples of C++ compiled applications and benchmarks (7zip - https://www.7-zip.org, GROMACS - https://www.gromacs.org/, ...) but nothing relevant enough to fully argue the proposal of this option.

In addition, the `vptr` sanitazer seems to be a more recent alternative, but I'm not familiar with it.

thesamesam: If I'm _really_ bored, I may consider trying it on a system-wide build of Gentoo for fun, but I don't plan on trying it. Someone else is welcome to though. I'd be interested in how it goes ;)

thomasnyman:
@Flo4152 I've opened a pull request against your branch to address the linter errors as well as some consistency issues with  the rest of the guide: https://github.com/Flo4152/wg-best-practices-os-developers/pull/1

Given the observation by @thesamesam in his https://github.com/ossf/wg-best-practices-os-developers/pull/485#discussion_r1594479772 regarding ABI breakage I think that is something that really warrants as **When to not use?** section I was [asking for in the original PR](https://github.com/ossf/wg-best-practices-os-developers/pull/440#discussion_r1546819433)

I am also wondering whether `-fsanitize=vptr` combined with the `-fsanitize-minimal-runtime` we were pointed to look at in https://github.com/ossf/wg-best-practices-os-developers/issues/326 would be a better options for production code. 

That said, I do think the material on the usage of `-fvtable-verify` in this PR is good and given that there are examples of users of this, even though it may not be fit for a general recommendation, I do think this would be good material for the guide as long as the caveats with the option are clearly communicated.

thesamesam:
My reservation is that I've genuinely never heard of anyone using this outside of those initial talks when it got merged into GCC. I'm sure users have existed, or do exist somewhere, but I've never heard from them and I don't ever see it mentioned in GCC bug reports or development discussions.

I'd feel a lot better about advertising this if:
* someone could ask around the GCC community (perhaps on the ML) what the consensus on the option is, and if it's suitable for the guide;
* someone could try build a bunch of software with it and see what happens (and make sure e.g. the compiler doesn't just ICE immediately)

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -993,6 +994,74 @@ The `-fexceptions` option is also needed for C code that needs to interoperate w
...
+
+- `std` - Built after shared libraries have been loaded and initialized, vtable data imply to process check at normal object initialization time.
+- `preinit` - Built before shared libraries have been loaded and initialized, vtable data enable check before shared libraries are loaded to process check at early execution time.
+- `none` - turns off the vtable checks.
+
+[^gccvtv]: Team GCC, [Program Instrumentation Options: `-fvtable-verify`](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtable-verify), GCC Manual, August 2013
+
+### Performance implications
+
+Caroline Tice played an important role in integrating the vtable verification (VTV) by submitting the commit of this feature in GCC 4.9[^gccvtvcommit]. She involved on articles that give information about performance penalty. She is describe performance penalty following 3 points:
```
thomasnyman, line 1019: There was feedback in the [C/C++ Compiler BP Guide Call in 2024-05-16](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.cobg2blj7ukq) that the "_performance implications section reads very differently than text elsewhere in the guide, should be reworded to be more consistent with other sections"_

This is perhaps partially because, unlikely the rest of the guide, which aims to acknowledge the original author's through appropriate references. This section seems to feature the author rather prominently. Maybe this paragraph could be shortened while still making sure the author's original work is included as a reference.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -993,6 +994,74 @@ The `-fexceptions` option is also needed for C code that needs to interoperate w
...
+
+- Call verification: Performance penalty between 5% and 10%.
+- Object permission changes: Between 400% and 700% slowdown for permission changes per object file, while a performance loss of 320 ms is noticeable for permission changes per binary.
+- Virtual function hashtable size: Storage of virtual function hashtable imply a big waste of space.
```
thomasnyman, line 1023: Another piece of feedback from the [C/C++ Compiler BP Guide Call in 2024-05-16](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.cobg2blj7ukq) was that _"performance numbers should be attributed to specific benchmarks, 5% impact in Chrome may be acceptable, but 5% impact on SPEC benchmarks would generally be undesirable"_

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -993,6 +994,74 @@ The `-fexceptions` option is also needed for C code that needs to interoperate w
...
+- `vtv_class_set_sizes.log` - The size of the vtable pointer sets for each class.
+
+[^Tice2012]: Tice, Caroline, [Improving Function Pointer Security for Virtual Method Dispatches](https://gcc.gnu.org/wiki/cauldron2012?action=AttachFile&do=get&target=cmtice.pdf#page=38), GNU Tools Cauldron, July 2012.
+[^Tice2014]: Tice, Caroline, [Enforcing Forward-Edge Control-Flow Integrity in GCC & LLVM](https://www.usenix.org/system/files/conference/usenixsecurity14/sec14-paper-tice.pdf), USENIX Security, August 2014.
+[^gccvtvcount]: Team GCC, [Program Instrumentation Options: `-fvtv-counts`](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtv-counts), GCC Manual, August 2013
+[^gccvtvcommit]: Tice, Caroline, [Commit the vtable verification feature](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=2077db1be5b18b94a91095a3fb380bbc4a81e61b), GCC GIT, August 2013
+
+#### When not to use?
+
+Using the `-fvtable-verify` flag can lead to ABI breaks.
```
thomasnyman, line 1039: Add a note here about the implications: _"Software that needs to interoperate with libraries built without `-fvtable-verify` should not enable this option._

thomasnyman (changes requested):

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
Flo4152: @thomasnyman @siddhesh The performance impact of the `-fvtable-verify` option is discussed in several articles.

In "Enforcing Forward-Edge Control-Flow Integrity in GCC & LLVM", a section is devoted to the performance impact of virtual table verification (7.1 VTV Performance - https://www.usenix.org/system/files/conference/usenixsecurity14/sec14-paper-tice.pdf#page=12). This article reports a performance impact on the "SPEC CPU2006 C++ benchmarks" of between 2.4% and 19.6% on omnetpp, astar and
xalancbmk. Four other benchmarks (povray, namd, soplex and dealII) showed no significant effect on performance. This article also focuses on reducing the performance penalty of the three affected benchmarks. They found that the performance penalty of xalancbmk fell from 19.2% to 10.8% using devirtualization and to 8.7% using static linking and devirtualization. They also noted that PGO optimization (flag -ripa) can reduce the number of virtual calls to be checked.

This information is presented in figures and graphs in the "Improving Function Pointer Security for Virtual Method Dispatches" slides on the GCC website (https://gcc.gnu.org/wiki/cauldron2012?action=AttachFile&do=get&target=cmtice.pdf) and in the slides relating to the article "Enforcing Forward-Edge Control-Flow Integrity in GCC & LLVM" (https://www.usenix.org/sites/default/files/conference/protected-files/sec14_slides_tice.pdf).

These numbers can add consistence to the section about performance impact. I also think that optimization that improving performance of code compiled with vtv also has a place in the section about performance impact (PGO, static linking and devirtualization) .

I will integrate these some of these numbers and optimizations into my pull request.

Regards,

david-a-wheeler:
These are really significant performance impacts, on a relatively narrowly focused countermeasure.

We definitely *should* document these options, including the performance impacts noted above. I don't think we should recommend using these options by default.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -68,7 +68,7 @@ Most programming languages prevent such defects by default. A few languages allo
...
 
-Modern, security-aware C and C++ software development practices, e.g., secure coding standards such as SEI CERT C[^CMU2016C] and C++[^CMU2016CPP], and program analysis aim to proactively avoid introducing memory errors (and other software defects) to applications. However, in practice completely eradicating memory errors in production C and C++ software has turned out to be near-impossible.
+Modern, security-aware C and C++ software development pracs, e.g., secure coding standards such as SEI CERT C[^CMU2016C] and C++[^CMU2016CPP], and program analysis aim to proactively avoid introducing memory errors (and other software defects) to applications. However, in prac completely eradicating memory errors in production C and C++ software has turned out to be near-impossible.
```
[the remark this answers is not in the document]
Flo4152: This unfortunate change is corrected in my last commit.

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
Flo4152: This caution is added

Flo4152:
@david-a-wheeler I describe performance impacts in my last change of "Performance implications" subsection. I think I summarized number from SPECS benchmarks. Do you think there's anything missing or should be completed?

I agree theses options should not recommend using these options by default. Do you mean these options should be moved from the "Recommended compiler options that enable run-time protection mechanisms." table?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
Flo4152: I summarized number from SPECS benchmarks in "Performance implications" subsection

david-a-wheeler:
Proposal - include text, but include at the end something like: “This option can have significant performance impact, while only countering a narrow type of vulnerability. Thus, we have not included this option as a recommended option in the current version of this guide.”

thesamesam:
I'll stop after this and won't insist further, but I have to ask:

> We definitely should document these options, including the performance impacts noted above. I don't think we should recommend using these options by default.

Why is this a _definitely_? Including things which we're saying people shouldn't use by default (and with a huge amount of caveats) is noise and it may end up reducing the effectiveness of the guide if it's a huge document with lots of irrelevant parts.

There are other options which have significant performance impact which _are_ worth mentioning because they're actively developed and tested at least. But to me, this particular one appears all-negative (https://github.com/ossf/wg-best-practices-os-developers/pull/485#issuecomment-2118983435). There are various options which are also not-great ideas which I could make PRs for. What makes this one different?

Isn't part of the goal here to focus on modern, useful, and sensible options?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -993,6 +994,82 @@ The `-fexceptions` option is also needed for C code that needs to interoperate w
...
+[^Tice2014slide]: Tice, Caroline, [Enforcing Forward-Edge Control-Flow Integrity in GCC & LLVM (slide)](https://www.usenix.org/sites/default/files/conference/protected-files/sec14_slides_tice.pdf#page=19), USENIX Security, August 2014
+[^Tice2014]: Tice, Caroline, [Enforcing Forward-Edge Control-Flow Integrity in GCC & LLVM](https://www.usenix.org/system/files/conference/usenixsecurity14/sec14-paper-tice.pdf#page=12) USENIX Security, August 2014
+[^gccvtvcount]: Team GCC, [Program Instrumentation Options: `-fvtv-counts`](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fvtv-counts), GCC Manual, August 2013
+[^SPECCPU2006]: Standard Performance Evaluation Corporation, [SPEC CPU 2006 benchmack](https://www.spec.org/cpu2006/) SPEC, August 2006
+[^SPECCPU2017]: Standard Performance Evaluation Corporation, [SPEC CPU 2017 benchmack](https://www.spec.org/cpu2017/) SPEC, June 2017
+
+#### When not to use?
+
+Using the `-fvtable-verify` flag can lead to ABI breaks.
+Software that needs to interoperate with libraries built without -fvtable-verify should not enable this option.
```
jwakely, line 1047: This includes the C++ standard library, and all system libraries. Those are typically not built with VTV enabled, so this option is only useful if you build your own GCC, and then self-build your entire software stack from the C++ std::lib upwards. In practice, that means it's useless.

I agree with Sam, this is just not good advice.

Flo4152: I agree with the difficulty to maintain binaries at Linux systems level. This agree this flag implying a lot of complication in case of system libraries (linking issues, etc...). 

First let me situate this flag in this guide. The description of this guide is about produce guide lines to build application with security mechanisms not only building a whole OS.

My proposal is mainly based on my experience in HPC environments where it is acceptable to offer different versions of compilers, HPC applications and related libraries on the same userspace (user context or application context).

HPC applications can be compiled and installed to accept custom library configurations using environment variables such as LD_PRELOAD or LD_LIBRARY_PATH, and PATH. This type of userspace can be maintained using tools such as "environment-modules" (https://modules.readthedocs.io/en/latest/), where application binaries and related libraries are built outside the system tree, in a hierarchy of custom folders.  Custom hierarchies are commonly used to offer applications with custom compilation configuration flags.  The use of a customized folder hierarchy offering the application and associated libraries compiled with VTV becomes reliable.

Let me expose a real-life use-case where the -fvtable-verify flag can be acceptable:
- "environment-modules" allow to provide custom build of GCC with –enable-vtable-verify flag is acceptable in a custom folder hierarchy without influencing other application or the whole system stability. The VTV enable GCC is only expose to userspace after the configuration of userspace using "environment-modules".
- All dependencies of application can be expose to userspace using custom folder hierarchy. Application folder hirarchy is maintained independently to the other application hierarchy that allow to ensure application reliability and maintainability without influencing other application or the whole system.
- The userspace configuration (using  environment-modules) is commonly dedicated to a unique high computing application a time. These dedicated environments exclude many issues related to interoperate with libraries built without `-fvtable-verify`.
- Performance is important in an HPC environment, and performance testing of an application is usually carried out in a validation environment before the application is exposed to the production user space. The figures from these benchmarks enable the manager of an application's file hierarchy to reject a version (optimizations flags can be used to reduce the impact on performance).  As you can see from Caroline Tice's articles, performance varies from benchmark to benchmark. Some applications will be rejected for performance issues, others accepted.

I understand that this type of environment can be unusual and may differ from a common system architecture (for example, a single version of a library, application and compiler maintained with the package manager), but it seems reliable to expose to userspace on-demande VTV-compatible compilers, applications and libraries in this way.

Regards,

jwakely: So this is for a specialized internal ecosystem, not a best practice for the wider open source ecosystem. So arguably off topic in this guide then.

Flo4152: Multi-version computer applications on custom folder hierarchies are common ecosystems. You can find many examples using this installation architecture, for example :
* https://docs.rcc.fsu.edu/hpc/environment-modules/
* https://docs.hpc.shef.ac.uk/en/latest/stanage/software/development/gcc.html

Moreover, most of project used on this king of architecture are opensource. A simple way to visualize this architecture is to compare it with installations prefixed to custom folders when configuring compilations (./configure --prefix=/path/) for the application and its dependencies. This practice is common in open-source ecosystems.

I'm not comparing it to a specialized internal ecosystem, but that's not an argument for including the -fvtable-verify flag in this guide. I'm not the right person to allow myself to justify the inclusion of this flag in this guide.

Flo4152:
@thomasnyman @david-a-wheeler There are arguments against adding this flag to your guide. This pull request was discussed at your zoom calls, and adding this flag seems to be in line with what was said at those calls.

Do you still feel the same way about adding this flag?

siddhesh:
It appears that the current state is the following:

* The `-fvtable-verify` option was implemented long ago with the idea that it may be useful, but never got any use in the Open Source ecosystem. Further, it appears that the feature is largely unmaintained since the feature implementors have since stopped contributing to GCC.
* The option in isolation breaks ABI, so it's unlikely that an individual project would be able to use it safely in any capacity.
* The option has a significant performance overhead

All of that sounds to me like this shouldn't go into the main guide, but should get mentioned in the Appendix as an option that was considered but rejected.

thomasnyman:
In the interest of trying to wrap this topic up after a lengthy discussion, I've proposed some text to add to the list of considred options for `-fvtable-verify`in #553.

Flo4152:
@thomasnyman @siddhesh I agree. I'm closing this request.

Flo4152:
Thank you all for your responses to this pull request. I appreciate the corrections, advice and help I received.

Regards,

Issue #490, opened 2024-05-16
Conversion of compiler option guide to machine readable format


mayank-ramnani:
In the future, there might be tools based on the OpenSSF compiler options guide.
This issue is a tracker to make the guide suitable to be converted into a machine readable format like json.
Tools such as the one I am working on (https://github.com/mayank-ramnani/temper) need the options in a machine readable format to do processing using them.

Currently, the guide has some inconsistencies which make it difficult to convert to a machine readable format.

Some issues in parsing:
1. Non-standard table format: 
- There are only a few rows in which the column “Compiler Flag” contains additional data other than the flag:
- Example: 
```
-D_FORTIFY_SOURCE=3
(requires -O1 or higher,
may require prepending -U_FORTIFY_SOURCE)
```
- Suggestion: Move the additional details to another column, maybe to description. Have a standard table format.

2. Non-standard version mentions.
- Suggestion: Mention till patch version (ex: gcc 2.95.3) instead of just the major version (ex: gcc 12)

**I will update this thread with more insights and can create a PR that implements the above.**

The script I created to scrape the webpage and convert it to json: doesn't give perfect results due to inconsistencies in the guide - https://github.com/mayank-ramnani/temper/blob/main/extra/scraper.py
- Example output from the script: https://github.com/mayank-ramnani/temper/blob/main/extra/scraped-db.json

mayank-ramnani:
Guide (HTML) can now be converted into machine readable format (JSON) using scraper.
Tables 1 and 2 (recommended compiler options) are extracted using the above mentioned scraper script.

PR #507, opened 2024-05-30
Standardize versions in the "supported since" columns to improve machine readability

mayank-ramnani:
As part of Issue https://github.com/ossf/wg-best-practices-os-developers/issues/490, change compiler and dependency versions to have until patch version.
The goal is to improve machine readability so scripts that extract compiler options from the webpage can easily extract the dependent versions.

For example, `gcc 12` -> `gcc 12.0.0`
Changes done for:
1. gcc
2. clang
3. binutils
4. libstdc++
5. libc++

Changes have only been done in tables, not in any of the paragraph text.

Github shows diff by lines. To see diff by characters (which is easier to read in this case), use - 
`git diff --color-words=.`


thomasnyman (approved):
lgtm

PR #514, opened 2024-05-31
Add Correctly Using Regular Expressions Guide (as 1 page)

david-a-wheeler:
This converts the Google document into markdown and puts in the docs/ directory. Eventually this should be split into 2 documents, but I think it's easier to do that separately.


david-a-wheeler:
Original document (now obsolete): https://docs.google.com/document/d/1Ors5T04Pgh3dcBfelbBrEBrvY3OKB7loUBUJPYBmmZw/edit

PR #536, opened 2024-06-12
Remove -D_LIBCPP_ASSERT from recommended options

wphuhn-intel:
In the Compiler Hardening Guide, `-D_LIBCPP_ASSERT` is listed as a recommended run-time options (Table 2 of current document), however it is also listed as an option that has been considered but excluded due to it being deprecated (Section 10 of current document.)

This pull request removes `-D_LIBCPP_ASSERT` from the list of recommended run-time options based on our discussion from the previous call.  If/how `-D_LIBCPP_ENABLE_HARDENED_MODE` should be recommended remains an open question.

The copyright date has also been updated.


thomasnyman (approved):
lgtm

PR #540, opened 2024-06-14
Labs: Handling Errors Lab

balteravishay:
Adding Lab for Handling Errors section

### DIFF
--- a/docs/labs/handling-errors.html
+++ b/docs/labs/handling-errors.html
@@ -1,0 +1,206 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="X-UA-Compatible" content="IE=edge">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link rel="stylesheet" href="https://best.openssf.org/assets/css/style.css">
+<link rel="stylesheet" href="checker.css">
+<script src="js-yaml.min.js"></script>
+<script src="checker.js"></script>
+<link rel="license" href="https://creativecommons.org/licenses/by/4.0/">
+
+<!-- See create_labs.md for how to create your own lab! -->
+
+<!-- Sample expected answer -->
+<script id="expected0" type="plain/text">
+throw new Error("Division by zero is not allowed");
+</script>
+<!--
+-->
+<script id="expected1" type="plain/text">
+  return a / b;
+</script>
+
+<script id="expected2" type="plain/text">
+  try {
+    const result = divide(10, 2);
+    console.log("Result:", result);
+  } catch (error) {
+    console.error("Error:", error.message);
+}
+   
+
+</script>
+
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+\s* throw new Error \( ("(.*?)" | '(.*?)' | `(.*?)`) \) ; \s*
+</script>
+<script id="correct1" type="plain/text">
+\s* return .\s* a \s* \/ \s* b \s* ; \s*
+</script>
+<script id="correct2" type="plain/text">
+  \s* try \s* { \s* 
+    const result  \s* =  \s* divide \( \s* 10  \s* ,  \s* 2  \s* \)  \s* ;
+    console.log \(  \s* ("Result:" | 'Result:' | `Result:`)  \s* ,  \s* result  \s* \)  \s* ;
+    \s* } \s* catch \(  \s* error  \s* \)  \s* {  \s*
+      \s* console.error \(  \s* ("Error:" | 'Error:' | `Error:`)  \s* ,  \s* error.message  \s* \)  \s* ;  \s*
+      \s* \} \s*
+  </script>
+
+<script id="info" type="application/yaml">
+---
+hints:
+- index: 0
+  present: "(Throw|THROW|New|NEW|error|ERROR)"
+  text: JavaScript is case-sensitive. use throw new Error(...).
+  examples:
+  - - "  Throw new Error(\"Division by zero is not allowed\");"
+  - - "  THROW new Error(\"Division by zero is not allowed\");"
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Huh! That is simpler than I was expecting... and that's a *good* thing! Thank you!!

Not everyone who does this lab will know the JavaScript syntax for exception handling, so we should probably add some hints to discuss that and/or include a little about JavaScript syntax in the description. I'll make a few specific comments.

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+  } catch (error) {
+    console.error("Error:", error.message);
+}
+   
+
+</script>
+
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+\s* throw new Error \( ("(.*?)" | '(.*?)' | `(.*?)`) \) ; \s*
```
david-a-wheeler, line 37: You don't need "?" (optionality) after ".*"; the "*" accepts "0 or more" characters, making characters optional in all three places.

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+</script>
+<script id="correct1" type="plain/text">
+\s* return .\s* a \s* \/ \s* b \s* ; \s*
```
david-a-wheeler, line 40: Why the "." after return?

david-a-wheeler: Also, you can just say:

`a / b`

One or more spaces is interpreted as `\s*`

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+</script>
+<script id="correct2" type="plain/text">
+  \s* try \s* { \s*
```
david-a-wheeler, line 43: Again, you don't need `\s*` between `try` and `{`.

However, I you need to escape the `{` here, it's a literal `{` character correct?

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+  examples:
+  - - "  Throw new Error(\"Division by zero is not allowed\");"
+  - - "  THROW new Error(\"Division by zero is not allowed\");"
+  - - "  throw New Error(\"Division by zero is not allowed\");"
+  - - "  throw NEW Error(\"Division by zero is not allowed\");"
+  - - "  throw new error(\"Division by zero is not allowed\");"
+  - - "  throw new ERROR(\"Division by zero is not allowed\");"
+- index: 0
+  absent: "throw"
+  text: Try using the throw keyword to raise an exception.
```
david-a-wheeler, line 66: I suggest being more specific since some learners won't necessarily have a solid knowledge of JavaScript syntax. E.g., append text something like this:

> E.g., throw new Error(...)

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+<p>
+<b>Please change the code below to leverage the exception handling mechanism and improve its robustness.</b>
+
+<p>
+<h2>Background</h2>
+<p>
+In this exercise, we'll demonstrate how to enhance the robustness of a JavaScript function by transitioning from a return code mechanism to exception handling. The primary goal is to understand how to handle errors more effectively, improving the readability and maintainability of the code.
+</p>
+Initially, the code uses a return code mechanism to handle potential errors. 
+The <tt>divide</tt> function returns an object indicating whether the operation was successful or if an error occurred. The caller must then check the success property and handle the result or error accordingly. This method works but can clutter the code with repetitive error-checking logic.
```
david-a-wheeler, line 134: Add:

"What's worse, it's easy for callers to ignore the error return code, which can result in surprising failures and security vulnerabilities".

We want to make sure they understand *why* they might make this change. It's also helpful to contrast this with languages like Rust & Haskell, which have a mechanism that at first *looks* like an error return, but *requires* thaat the caller deal with the error case.

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+\s* throw new Error \( ("(.*?)" | '(.*?)' | `(.*?)`) \) ; \s*
+</script>
+<script id="correct1" type="plain/text">
+\s* return .\s* a \s* \/ \s* b \s* ; \s*
+</script>
+<script id="correct2" type="plain/text">
+  \s* try \s* { \s* 
+    const result  \s* =  \s* divide \( \s* 10  \s* ,  \s* 2  \s* \)  \s* ;
```
david-a-wheeler, line 44: Again, you don't need `\s*` between spaces, spaces are interpreted as `\s*`

(The system *will* optimize these away, so they aren't *wrong*, but they do make it harder to see the pattern of what is correct.)

docs/labs/handling-errors.html
```diff
@@ -0,0 +1,206 @@
...
+
+<p>
+<h2>Task</h2>
+<p>
+<b>Please change the code below to leverage the exception handling mechanism and improve its robustness.</b>
+
+<p>
+<h2>Background</h2>
+<p>
+In this exercise, we'll demonstrate how to enhance the robustness of a JavaScript function by transitioning from a return code mechanism to exception handling. The primary goal is to understand how to handle errors more effectively, improving the readability and maintainability of the code.
```
david-a-wheeler, line 131: This is a remarkably simple & straightforward lab exercise, and that's great!

balteravishay: thanks! glad you apperciate that!

balteravishay:
Thanks @david-a-wheeler , fixed your comments.

david-a-wheeler:
Awesome! Thanks so much! I've merged it in.

PR #559, opened 2024-07-03
Add lesson learned from polyfill.io

david-a-wheeler:


docs/Concise-Guide-for-Developing-More-Secure-Software.md
```diff
@@ -33,5 +33,6 @@ Here is a concise guide for all software developers for secure software developm
 24. **Continuously improve**. Improve scores, look for tips, & apply as appropriate.
 25. **Manage succession**. Have clear governance & work to add active, trustworthy maintainer(s).
 26. **Prefer memory-safe languages**. Many vulnerabilities involve memory safety. Where practical, use memory-safe programming languages (most are) and keep memory safety enabled. Otherwise, use mechanisms like extra tools and peer review to reduce risk.
+27. **Ensure production websites only load from your own domains**. _Linking_ to other domains is fine, but don't load information from other domains such as JavaScript, CSS, and images. If you do, that domain's subversion might also subvert your site. See the [subverted polyfill.io revealed in 2024](https://blog.qualys.com/vulnerabilities-threat-research/2024/06/28/polyfill-io-supply-chain-attack).
```
ctcpip, line 36:
```suggestion
27. **Ensure production websites only load assets from your own domains**. _Linking_ to other domains is fine, but don't load assets from other domains such as JavaScript, CSS, and media. If you do, that domain's subversion might also subvert your site. See the [subverted polyfill.io revealed in 2024](https://blog.qualys.com/vulnerabilities-threat-research/2024/06/28/polyfill-io-supply-chain-attack).
```

also, should the text reference the term supply chain attack or similar?

ctcpip, line 36: should the text reference the term supply chain attack or similar?

david-a-wheeler: I like the reword, thanks!

ctcpip:
> I wouldn't say "supply chain" on just this item. There are many supply chain attacks, and this is only one kind. If we single this out with the term "supply chain", I fear people might think they're identical.

I actually backtracked on that and had removed my comment, but I guess there was a race condition :)

ctcpip (approved):

david-a-wheeler:
@ctcpip - no problem! I merged in your improvement, & made a few more tweaks. I think it's better, thank you!

ljharb (approved):
i thought we all learned this 10-20 years ago, but apparently not :-p

david-a-wheeler:
[ljharb](https://github.com/ljharb) said:
> i thought we all learned this 10-20 years ago, but apparently not :-p

Hehe. I think the problem is that new people keep showing up, and instead of learning from the past, they presume that the past has nothing to teach them. The past has very brutal ways to show otherwise :-(.

SecurityCRob (approved):
+1

drusso-rh:
+1

Issue #531, opened 2024-06-05
Python guide: Doc2GitHub, moving code from an Ericsson internal confluence to this GitHub space.


myteron:
There are around 40 rules on and Ericsson internal confluence that have approval by Ericsson Opensource group to be published. 
Some of the text and code requires refactoring. 
This work can only be done by Ericsson employees.
Once all docs are made available in GitHub we will have GitHub as the main source for these documents and stop using Confluence for the Python secure coding.

myteron:
@SecurityCRob I wonder what the best handling is. I could start ~~striking through~~ the CWE's we have processed but it overall appears to me that an "issue" is to small for what we are trying to do here. It seems that a project is the next level up to an issues but never used that.
Not sure how well a milestone would work for this.

myteron:
52 rules made it into GitHub.
3 Rules had issues, so we created GitHub issues:
* CWE-116: Prevent XML Injection, created issue on GitHub #986
* CWE-369, Divide by Zero, created issue on GitHub: #988
* XXX-005: Consider hash-based integrity verification of byte code files against their source code files, created issue on GitHub #711

Closing this.

PR #566, opened 2024-07-29
PyDoc2GitHub CWE-426

tommcd:
Moving CWE-426 from confluence as part of #531

Note: existing folder XXX-005 should have been CWE-426 so it has been renamed.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-426/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-426/README.md
@@ -1,0 +1,94 @@
+# CWE-426: Untrusted Search Path
+
+In an environment where an untrusted or less trusted entity can modify the environment variables, consider validating hash-based byte code [Python 2023 Command line and environment].
+
+Python source code `.py`  files need to be converted into "byte code" `.pyc` or `.pyo` in memory or in a filesystem `__pycache__` before running on the Python Virtual Machine (PVM) [Dec 2009 PEP 3147].
+Python 3.8 [Dec 2009 PEP 3147] also has a backward compatibility mode supporting delivering only byte code.
+
+Python 3.8 introduced the option to customize the `__pychache__` folder via `-X pycache_prefix=PATH`, or the `PYTHONPYCACHEPREFIX` environment variable. An attacker may manipulate the `PYTHONPYCACHEPREFIX` or `PYTHONPATH` to inject their code that can go unnoticed without hash-based verification. Without `--check-hash-based-pycs` Python only compares the byte code against the source code via time-stamp [Python 2023 The import system] potentially allowing an attack.
+
+Python 2.6 also introduced the ability to stop Python from writing "byte code" files via `-B` flag or `PYTHONDONTWRITEBYTECODE=1` environment variable is another possibility that does not guarantee full protection.
+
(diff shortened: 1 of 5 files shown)


tommcd:
Hi @gkunz,
I wanted to add  Labels "Product: Python Hardening Guide" to this PR, but I don't see any way to do so.
The 'Labels' section on the right side of this page is not editable for me.
I assume I don't have the required permissions?

gkunz:
> Hi @gkunz, I wanted to add Labels "Product: Python Hardening Guide" to this PR, but I don't see any way to do so. The 'Labels' section on the right side of this page is not editable for me. I assume I don't have the required permissions?

Hi @tommcd: yes, that seems to be the case. I just applied the label.

docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-426/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## Exceptions
+
+**ENV-4P-EX0:** Untrusted entities are not able to change environmental variables or any Python files.
+
+## Risk Assessment
+
+An attacker with access to the environment and filesystem can allow full control over what a Python script can run.
+
```
myteron, line 59: Remove Risk Assessment

gkunz:
I tried to validate the example, but I don't fully understand the effect of the `--check-hash-based-pycs` flag. To my understanding, it should avoid some sort of cache poisoning, but I wasn't able to make the example work for me with regards to this option.

To be precise: the non-compliant example works as expected, but that's due to the `-I` flag.

Is my understanding correct that one should try to construct an example in which the cached file include different code than the original source? I tried to create two different cache files (one for the server, one for the print out) and replace one with the other, but that didn't work for me. Is that the intention?

SecurityCRob (approved):
thank you

tommcd:
> I tried to validate the example, but I don't fully understand the effect of the `--check-hash-based-pycs` flag. To my understanding, it should avoid some sort of cache poisoning, but I wasn't able to make the example work for me with regards to this option.
> 
> To be precise: the non-compliant example works as expected, but that's due to the `-I` flag.
> 
> Is my understanding correct that one should try to construct an example in which the cached file include different code than the original source? I tried to create two different cache files (one for the server, one for the print out) and replace one with the other, but that didn't work for me. Is that the intention?

I believe you are correct. For now I reworded the description to make it clearer that -I address the exploit but `--check-hash-based-pycs always` adds additional security. I'll discuss with Helge to see if another or berrer example can be used.

tommcd:
Hi @gkunz,
I'm not sure how to  fix the following:

"Changes requested
1 review requesting changes and 1 approving review by reviewers with write access. 

Merging is blocked
Merging can be performed automatically once the requested changes are addressed."

As far as I can see the requested changes were already resolved?

gkunz:
> Hi @gkunz, I'm not sure how to fix the following:
> 
> "Changes requested 1 review requesting changes and 1 approving review by reviewers with write access.
> 
> Merging is blocked Merging can be performed automatically once the requested changes are addressed."
> 
> As far as I can see the requested changes were already resolved?

@tommcd sorry, my bad. I'll update my review. Thanks for addressing my comments.

gkunz (approved):

PR #571, opened 2024-08-08
Add -Wl,--as-needed and -Wl, --no-copy-dt-needed-entries to C and C++ Compiler Hardening Guide

### LINKED ISSUE #510: Consider `Wl,--as-needed` and `-Wl,--no-copy-dt-needed-entries` for C and C++ Compiler Hardening Guide
The `Wl,--as-needed` flag tells the linker to link in the produced binary only the libraries containing symbols actually used by the binary itself. This can positively impact security by reducing the attack surface of the produced binaries in terms of, e.g., gadgets available for code reuse attacks. 

The `Wl,--no-copy-dt-needed-entries` affects the resolution of symbols in dynamic libraries by limiting the symbol resolution to the shared library expected to contain a specific symbol. This is the default in `ld`. Potentially the inverse, `Wl,--copy-dt-needed-entries` could be added as a discouraged option as that allows symbols to be resolved implicitly linked libraries in application code.

References:
- https://wiki.gentoo.org/wiki/Project:Quality_Assurance/As-needed
- https://bugzilla.redhat.com/show_bug.cgi?id=1292230

thomasnyman:
Fixes #510

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -27,7 +27,8 @@ When compiling C or C++ code on compilers such as GCC and clang, turn on these f
 -fstrict-flex-arrays=3 \
 -fstack-clash-protection -fstack-protector-strong \
 -Wl,-z,nodlopen -Wl,-z,noexecstack \
--Wl,-z,relro -Wl,-z,now
+-Wl,-z,relro -Wl,-z,now \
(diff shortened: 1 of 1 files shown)


thomasnyman:
Appeased to linter and ensured the Binutils version numbers match the convention in eed5267b30544e3861dc31ca791c1acd0d82642d.

thomasnyman:
@siddhesh Tagging you here since you indicated in the [C/C++ Compiler BP Guide on 2024-06-13](https://docs.google.com/document/d/1JY8FREBPCUUFpuv7-4B9EjeS2MLDpel0dbG5DFWrTns/edit#bookmark=id.lqqttg8d7hfr) that you might be interested to contribute or comment on this.

PR #580, opened 2024-08-13
Minor cleanup of lab handling-errors

david-a-wheeler:


david-a-wheeler:
The term "error" is used in two different ways in the lab, which is confusing. Let's separate those uses.

david-a-wheeler:
I want to merge this asap. Having *two* different uses for the word "error" - on the same line! - creates a really confusing exercise.

PR #599, opened 2024-08-25
Shell injection hints3

david-a-wheeler:


david-a-wheeler:
This improves the shell injection hints further. Clearly there are questions on this lab, so we want to provide some extra helps.

PR #609, opened 2024-09-04
Add notes about the "hint" and "give up" buttons in labs

david-a-wheeler:
Per discussion in education SIG meeting 2024-09-04, add a note in each lab about the "hint" and "give up" buttons.


david-a-wheeler:
This was already agreed to (in principle) in the education SIG meeting today, so I'm merging. We can make small tweaks later.

PR #618, opened 2024-09-19
add sending output format strings lab with python string format

jasinner:
Uses the python string format to demonstrate the [Format Strings and Templates ](https://github.com/ossf/secure-sw-dev-fundamentals/blob/main/secure_software_development_fundamentals.md#format-strings-and-templates) section of course.

This example came from [Be Careful with Python's New-Style String Format](https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/) by Armin Ronacher and was produced in collaboration with @david-a-wheeler

### DIFF
--- a/docs/labs/format-strings.html
+++ b/docs/labs/format-strings.html
@@ -1,0 +1,136 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="X-UA-Compatible" content="IE=edge">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link rel="stylesheet" href="https://best.openssf.org/assets/css/style.css">
+<link rel="stylesheet" href="checker.css">
+<script src="js-yaml.min.js"></script>
+<script src="checker.js"></script>
+<link rel="license" href="https://creativecommons.org/licenses/by/4.0/">
+
+
+<script id="expected0" type="plain/text">
+    return '{format_string}, {event.level}'.format(format_string=user_input, event=event)
+</script>
+
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+\s*return\s'\{format_string\},\s\{event\.level\}'\.format\(format_string=user_input,\sevent=event\)\s*
+</script>
+
+<script id="info" type="application/yaml">
+---
+- absent: "user_input"
+  text: Make sure the user_input is included in the replacement fields passed to the format function
+  examples:
+  - - "'some format'.format(user_input=user_input)"
+  - - "'some format'.format(user_input)"
+- absent: "event"
+  text: Make sure the event is included in the replacement fields passed to the format function
+  examples:
+  - - "'some format'.format(user_input=user_input, event=event)"
+  - - "'some format'.format(user_input, event)"
+- absent: \{(0|format_string)\}
+  text: Make sure the braces for the format_string replacement field are in the format string
+  examples:
+  - - "'{event}'.format(format_string=user_input)"
+  - - "'{0}.format(user_input)"
+- absent: \{(1|event)\}
+  text: Make sure the braces for the event replacement field are in the format string
+  examples:
+  - - "'{format_string}, {event.level}'.format(format_string=user_input, event=event)"
+  - - "'{0}, {1.level}'.format(user_input, event)"
+- absent: \{(1\.level|event\.level)\}
+  text: Make sure the level attribute of the event replacement field is in the format string
+  examples:
+  - - "'{format_string}, {event.level}'.format(format_string=user_input, event=event)"
+  - - "'{0}, {1.level}'.format(user_input, event)"
+# debug: true
+</script>
+</head>
+<body>
+<!-- For GitHub Pages formatting: -->
+<div class="container-lg px-3 my-5 markdown-body">
+<h1>Lab Exercise Format Strings and Templates</h1>
+<p>
+This is a lab exercise on developing secure software.
+For more information, see the <a href="introduction.html" target="_blank">introduction to
+the labs</a>.
+
+<p>
+<h2>Task</h2>
+<p>
+<b>Practice using string templates in a secure way.</b>
+
+<p>
+<h2>Background</h2>
+<p>
+In this exercise, we'll adjust our string formatting so that it doesn't allow a user to control
+the <a href="https://docs.python.org/3/tutorial/inputoutput.html#the-string-format-method"><tt>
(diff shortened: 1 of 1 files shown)


jasinner:
I'm having trouble getting the hints to work. Appreciate some help with that @david-a-wheeler

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+Use the “hint” and “give up” buttons if necessary.
+
+<p>
+<h2>Interactive Lab (<span id="grade"></span>)</h2>
+<p>
+<form id="lab">
+<pre><code
+> # Application configuration which should be kept secret from a user
+CONFIG = {
+    'SECRET_KEY': 'super secret key'
```
david-a-wheeler, line 105: Don't do that! That would be a secret key embedded in source code, which is *itself* a vulnerability.

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+<b>Practice using string templates in a secure way.</b>
+
+<p>
+<h2>Background</h2>
+<p>
+In this exercise, we'll adjust our string formatting so that it doesn't allow a user to control
+the <a href="https://docs.python.org/3/tutorial/inputoutput.html#the-string-format-method"><tt>
+format string</tt></a>. If a user can control the <tt>format string</tt> they can access
+variables which they shouldn't. Particularly if those variable's values can be returned to the user
+as output, it could lead to information disclosure beyond what was intended by the developer.
```
david-a-wheeler, line 73: I'd add something like this:

> In some cases the results can be even more serious, depending on the situation and programming language used. If a similar problem happened in C's printf function, it would often enable an attacker to completely control the program.

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+
+<p>
+Please change the code below so the string formatting cannot disclose arbitrary
+program values. The server-side program is written in Python and allows a user to specify a
+<tt>format string</tt> to control the output format of an event.
+
+<p>
+You could adjust the program so that it only formats the event, and does not include any user input,
+However it is considered safe to include user input in the output as long as they cannot control
+the <tt>format string</tt> itself.
```
david-a-wheeler, line 87: That's WAY too strong a claim. Just say that allowing arbitrary user control of format strings is dangerous.

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+
+<p>
+Adjust the value returned by the <tt>format_event</tt> function so the the user controlled
+<tt>user_input</tt> variable is only used as a
+<a href="https://docs.python.org/3/library/string.html#format-string-syntax"><tt>replacement field</tt></a>
+and is not used as the <tt>format string</tt>.
```
david-a-wheeler, line 93: Huh? This doesn't look secure.

This still allows for user control of the field to be shown. Depending on the object being passed to format, that can *still* leak secret data that isn't supposed to be shared. It *might* be okay in context, but I think the lab should focus on the "key point" which is that format strings can be dangerous, and if you don't HAVE to have them controlled by the user, don't.

david-a-wheeler: The [Be Careful with Python's New-Style String Format](https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/) article uses "the somewhat undocumented internals to change the behavior" in an attempt to sandbox things. That's a complex and dangerous way to resolve security issues. Sure, sometimes you must do something quirky like that, but there's a simple one: don't do it. If you only get format strings from trusted sources, the problem disappears. The article mentions translators, but usually translators are a small number of trusted people, very different from accepting arbitrary inputs from attackers.

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link rel="stylesheet" href="https://best.openssf.org/assets/css/style.css">
+<link rel="stylesheet" href="checker.css">
+<script src="js-yaml.min.js"></script>
+<script src="checker.js"></script>
+<link rel="license" href="https://creativecommons.org/licenses/by/4.0/">
+
+
+<script id="expected0" type="plain/text">
+    return '{format_string}, {event.level}'.format(format_string=user_input, event=event)
```
david-a-wheeler, line 14: This still allows untrusted user input as a format string, and that's what we are telling them to avoid. Also, while "event=event" is Pythonic, it's really unclear to non-experts in Python, because they don't know which "event" is being used in the format string.

I'd expect something like this as an answer:

~~~~
 return 'Event level = {event_level}'.format(event_level=event.level)
~~~~

docs/labs/format-strings.html
```diff
@@ -0,0 +1,136 @@
...
+</script>
+
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+\s*return\s'\{format_string\},\s\{event\.level\}'\.format\(format_string=user_input,\sevent=event\)\s*
+</script>
+
+<script id="info" type="application/yaml">
+---
+- absent: "user_input"
```
david-a-wheeler, line 24: You're missing this line after `---`:

~~~~
hints:
~~~~

We support several different keys in the info, "hints" is one of them.

Once you do, you'll find a lot of the tests fail. You'll need the tests to pass :-).

david-a-wheeler:
Thanks so much for the update! I provided a few comments. Can you please do a DCO signoff?

david-a-wheeler:
Hi - you're so close! Can you please finish this soon? Thanks so much!

david-a-wheeler:
> I'm having trouble getting the hints to work. Appreciate some help with that @david-a-wheeler

You're missing the `hints:` field identifier. You should see something like this:

~~~~
<script id="info" type="application/yaml">
---
hints:
- absent: "user_input"
~~~~

david-a-wheeler:
Can you try to wrap this up sooner if you can? We're trying to get a number of the labs out-the-door :-).

Also: I had assigned "Debug and Assertion" to you as well. I thinking of unassigning until you complete this one, and once you're done with this, let me know which available ones you'd like to do. Pre-assigning more than one was probably a bad idea, sorry about that.

david-a-wheeler:
Hi, I haven't heard from you in 2 weeks, and we need to get moving. I *do* appreciate the work you've done so far. I plan to complete your lab myself and merge it, and do the debug lab myself instead. Thanks!

david-a-wheeler:
More needs fixing, but it's easier to merge first & create separate fixes.

PR #627, opened 2024-09-22
Add insecure deserialization lab

camilaavilarinho:
Adding the lab for [Insecure Deserialization](https://github.com/ossf/secure-sw-dev-fundamentals/blob/main/secure_software_development_fundamentals.md#insecure-deserialization)

### DIFF
--- a/docs/labs/insecure-deserialization.html
+++ b/docs/labs/insecure-deserialization.html
@@ -23,10 +23,10 @@
 
 <!-- Full pattern of correct answer -->
 <script id="correct0" type="plain/text">
- const\s+data = JSON\.parse\(base64Decoded\)\; \s*
+ const\s+data = JSON \. parse \( base64Decoded \) \; \s*
 </script>
 <script id="correct1" type="plain/text">
- if \(data\.username && typeof\s+data\.username == ('string'|"string"|`string`) && data\.username\.length < 20\) \{ \s*
+ if \( data \. username && typeof\s+data \. username == ('string'|"string"|`string`) && data \. username \. length < 20 \) \{ \s*
 </script>
 
 <script id="info" type="application/yaml">
@@ -34,6 +34,21 @@
 hints:
 - present: "json"
   text: the JSON built-in global object is witten in uppercase.
+- absent: data.username
+  index: 1
+  text: Check if the data object has a property called username. You can do this by referencing data.username.
+- absent: \&\&
+  index: 1
+  text: To combine multiple conditions in JavaScript use &&. This operator means 'and', so both conditions must be true for the entire statement to pass.
+- absent: typeof
+  index: 1
+  text: Use typeof to check the type of the operand's value.
+- present: typeof data.username == 'String'
+  index: 1
+  text: When using typeof, JavaScript expects "string" all lowercase.
+- absent: length
+  index: 1
+  text: check if the length of the string is smaller than 20 characters.
 
 # debug: true
 </script>
--- a/docs/labs/insecure-deserialization.html
+++ b/docs/labs/insecure-deserialization.html
@@ -1,0 +1,136 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="X-UA-Compatible" content="IE=edge">
+<meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


docs/labs/insecure-deserialization.html
```diff
@@ -0,0 +1,136 @@
...
+</script>
+<!--
+-->
+<script id="expected1" type="plain/text">
+  if (data.username && typeof data.username == 'string' && data.username.length < 20) {
+</script>
+
+<!-- Full pattern of correct answer -->
+<script id="correct0" type="plain/text">
+ const\s+data = JSON\.parse\(base64Decoded\)\; \s*
```
david-a-wheeler, line 26: Overall correct0 looks good. There should be spaces around `\.` and `(` and `)` and `\;` because in JavaScript whitespace is optional around each.

docs/labs/insecure-deserialization.html
```diff
@@ -0,0 +1,136 @@
...
+</script>
+<script id="correct1" type="plain/text">
+ if \(data\.username && typeof\s+data\.username == ('string'|"string"|`string`) && data\.username\.length < 20\) \{ \s*
```
david-a-wheeler, line 29: Again, in correct1 I think there should be spaces around `(` and `\.` and `\)` because whitespace is optional around them.

I think after typeof you should allow the following parameter to be surrounded by a parenthesized pair, that is, `typeof ( \( data \. username \) | data \. username )` - treating typeof as a function/method call isn't insane, and JavaScript would accept it.

docs/labs/insecure-deserialization.html
```diff
@@ -0,0 +1,136 @@
...
+</script>
+
+<script id="info" type="application/yaml">
+---
+hints:
+- present: "json"
+  text: the JSON built-in global object is witten in uppercase.
```
david-a-wheeler, line 36: Not everyone who takes the course will know JavaScript, and in particular, they may not be familiar with JavaScript syntax.

I suggest creating several hints for someone who's trying to fill in part1, but is getting stuck on one of the parts. Break down each part of the condition:

> data.username && typeof data.username == 'string' && data.username.length < 20) {

So if they're missing "data.username", suggest that as a way to detect if it's there. Then, if they're missing the typeof, suggest that. Many won't know if the result is 'String' or 'string', help them there. And so on. Adding a few hint patterns can really help someone who understands the concept but can't quite get the syntax right.

camilaavilarinho: Thanks for reviewing it! I'll add your suggestions.  

I tried to add more hints, but when I use "absent" for one of the parts, it still considers absent if the value is not in both parts. For example if I add "absent: length..." and then fill part 2 with length, the hint continues to show because length is not in part 1 as well. Am I doing something wrong?
Maybe is better if I change to have just one input area for both parts?

david-a-wheeler: By default it only looks at the first part, which is **index 0**. If you want it to look at a different part, you need to give it an **index: 1** or whatever the index is. You can see examples here:

* https://best.openssf.org/labs/sql-injection.html
* https://best.openssf.org/labs/regex1.html

If you want to check *both* places, currently you have to repeat the rule. If that happens a lot I guess we could add modify checker.js so you could provide `index: LIST`, but let's see if that happens often first...!

david-a-wheeler:
Hi - I'm really eager to get your lab "over the hump" and into the course. Will you be making these changes by Oct 9? Otherwise, I think I'll merge this in, and make a few fix up tweaks myself. I want to ensure people can *use* the wonderful material you've developed. It's so close, I want to get this completed and released.

Thank you again!

camilaavilarinho:
> Hi - I'm really eager to get your lab "over the hump" and into the course. Will you be making these changes by Oct 9? Otherwise, I think I'll merge this in, and make a few fix up tweaks myself. I want to ensure people can _use_ the wonderful material you've developed. It's so close, I want to get this completed and released.
> 
> Thank you again!

Sorry the delay!

david-a-wheeler:
As noted earlier, I'm going to merge this draft lab in, and I'll make proposed tweaks as a separate PR to address the review comments above.

THANK YOU for all your work!!

PR #629, opened 2024-09-23
Fix "Correctly Using Regex" table for Python and Ruby

david-a-wheeler:
The detailed rationale explained why "$" is permissive in Python3 and Ruby, but the roll-up table is wrong (!). Fix the table, and provide source code for verifying it.


SecurityCRob (approved):
discussed today in WG call  +1

PR #638, opened 2024-10-02
Shell argument injection tweaks

david-a-wheeler:


david-a-wheeler:
@lirantal - Thanks so much for the lab. I made a number of tweaks, e.g., made it more general, removed some unnecessary text, and added a lot of hints. Some developers who take the course may not know JavaScript (or might not know it well), so hints for common mistakes can be a real help.

Please let me know what you think. If this is okay, let's declare this done.

lirantal:
Thanks will take a look 👀

docs/labs/argument-injection.html
```diff
@@ -14,60 +14,123 @@
...
+    \[ ['"`]blame
+  text: >
+    Pass the arguments as an array, like this:
+    `execFile('git', ['blame', ...])`.
+- present: |-
+    --
+  absent: |-
+    ['"\`]--['"`]
+  text: >
+    To pass `--` you need to pass it as a constant string. Typically this
```
lirantal, line 60:
```suggestion
    To pass `--` you need to pass it as a literal string. Typically this
```

lirantal (approved):
Looks excellent!
I proposed a small change to the wording but can land otherwise.

lirantal:
Have to admit, the regex parts for the hint checks is kinda annoying but you seem to master that so maybe you'll help me with the next lab too :-)

david-a-wheeler:
Python is the hardest for the regex. check.js has a mechanism for changing how it interprets regexes, but in this case, it's probably best to use it the way it is.

david-a-wheeler:
We need to add the new lab to the course. Issue to remember that is here: https://github.com/ossf/secure-sw-dev-fundamentals/issues/175

PR #666, opened 2024-10-14
Fix lab format-strings

david-a-wheeler:
Add "hints:" (without the field name, it has no contents).

Remove an example of a password in source code.
That is a vulnerability all by itself, we do NOT want to show how to write *vulnerable* code.

### DIFF
--- a/docs/labs/format-strings.html
+++ b/docs/labs/format-strings.html
@@ -21,6 +21,7 @@
 
 <script id="info" type="application/yaml">
 ---
+hints:
 - absent: "user_input"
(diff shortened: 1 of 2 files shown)


david-a-wheeler:
[jasinner](https://github.com/jasinner) - thank you for your earlier work! I think it'd be better if we simplified the actual lab, so we just switch to constant strings. We don't need to focus on how to attack the wrong code in the lab itself. I changed it to merely hint at an attack, and checked in full Python source code that demonstrates the details of the attack for those who want to verify it.

PR #670, opened 2024-10-17
Discourage use of sanitizers for SUID binaries

### LINKED ISSUE #661: Sanitizers need a bigger caveat with suid binaries at least
We should make clearer that sanitizers should not be used in production, at least not without minimal runtime (see https://github.com/ossf/wg-best-practices-os-developers/issues/326). Needs to be checked whether GCC's just-trap option for sanitizers is similar enough.

Using sanitizers in production with suid binaries can introduce privilege escalation vulnerabilities if a runtime is used as `ASAN_OPTIONS` or `UBSAN_OPTIONS` (possibly others) allow specifying an arbitrary log file path. See https://www.openwall.com/lists/oss-security/2016/02/17/9.

thomasnyman:
Fixes #661

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -1133,6 +1133,10 @@ While more efficient compared to dynamic analysis, sanitizers are still prohibit
 
 As with all testing practices, sanitizers cannot absolutely prove the absence of bugs. However, when used appropriately and regularly they can help in identifying latent memory, concurrency, and undefined behavior-related bugs which may be difficult to pinpoint.
 
+Sanitizers should not be used for hardening in production environments, particularly for Set User ID (SUID) binaries, as they expose operational parameters via environmental variables which can be manipulated to clobber root-owned files and privilege escalation[^Nagy2016].
+
+[^Nagy2016]: Nagy , Szabolcs, [Address Sanitizer local root](https://www.openwall.com/lists/oss-security/2016/02/17/9), Openwall mailing list, 2016-02-16.
+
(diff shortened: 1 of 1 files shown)


docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -1133,6 +1133,10 @@ While more efficient compared to dynamic analysis, sanitizers are still prohibit
 
 As with all testing practices, sanitizers cannot absolutely prove the absence of bugs. However, when used appropriately and regularly they can help in identifying latent memory, concurrency, and undefined behavior-related bugs which may be difficult to pinpoint.
 
+Sanitizers should not be used for hardening in production environments, particularly for Set User ID (SUID) binaries, as they expose operational parameters via environmental variables which can be manipulated to clobber root-owned files and privilege escalation[^Nagy2016].
+
+[^Nagy2016]: Nagy , Szabolcs, [Address Sanitizer local root](https://www.openwall.com/lists/oss-security/2016/02/17/9), Openwall mailing list, 2026-02-16.
```
thesamesam, line 1138:
```suggestion
[^Nagy2016]: Nagy , Szabolcs, [Address Sanitizer local root](https://www.openwall.com/lists/oss-security/2016/02/17/9), Openwall mailing list, 2016-02-16.
```

thesamesam (approved):
Thank you!

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
thomasnyman: Thanks! Good catch! Fixed in 91080eba07395b12b819cd1dc92648f9158921d9.

PR #675, opened 2024-10-21
Add to background more on *why* C and C++ will keep existing

david-a-wheeler:
This includes several citations.


thomasnyman:
Suggested some edits to improve readability in 90e2a74b874e31ac504a0b8a90712a47716f121f

Content-wise, I tried to put the monetary estimate into context of in terms of scale.

This also fix linter warnings due to two adjacent references. It appears this is a limitation in the linter rather than a real Markdown problem, but to work around that I've moved the reference to the The Register article to another sentence.

PR #681, opened 2024-10-31
Fix fallthrough example macro to work with very old gcc versions

garazdawi:
To deal with old gcc versions that do not have `__has_attribute`, the gcc manual suggest using this form. 

See https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html


thomasnyman (approved):
lgtm

PR #690, opened 2024-11-01
pySCG adding missing CVEs for 755 and 532

myteron:
pySCG main readme table was missing the CVEs for 755 and 532 as part of #531 , added now.


BartyBoi1128 (approved):
Looks good to me!

PR #699, opened 2024-12-03
Add cross-reference to Java regex rationale. Fixes #548

### LINKED ISSUE #548: clarification needed in RegEx guide
in https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Correctly-Using-Regular-Expressions.md, the following text appears in the Java row: `“\z”; “$” works but some documents conflict`

What does this mean exactly?  Let's add some information/context to this in the guide.  Alternately, remove the qualification ("but...") if it's not needed.  My understanding is that Java `$` will only match the end of the input (when not in multiline mode; equivalent to `\z`) and is therefore safe to use.

cc @david-a-wheeler

david-a-wheeler:


docs/Correctly-Using-Regular-Expressions.md
```diff
@@ -60,7 +60,7 @@ Platform
    </td>
    <td>“^” or “\A”
    </td>
-   <td>“\z”; “$” works but some documents conflict
+   <td>“\z”; [“$” works but some documents conflict](./Correctly-Using-Regular-Expressions-Rationale#java)
```
gkunz, line 63: It seems that this link does not render correctly because it is within a HTML table cell. Why is this a HTML table in the first place?

gkunz: @david-a-wheeler, I took the liberty to convert the HTML table into a markdown table so that the link actually works. Let me know if using a markdown table is controversial.

david-a-wheeler: It's a markdown file, so using markdown tables isn't controversial to me. I think earlier on we used HTML table capabilities that aren't in Markdown (even GitHub flavored markdown). If we don't need those capabilities, then that's fine.

gkunz: ok, thanks. I propose to go with the markdown table then.

gkunz (approved):

PR #706, opened 2024-12-12
Clarify compiler options hardening limitations when linking to pre-built artifacts 

### LINKED ISSUE #705: Clarification needed for projects linking against prebuilt static libraries
From @jduck via the OpenSSF Slack:

> Recently I discovered that taking dependency on third party proprietary (prebuilt) static libraries can lead to a false sense of security. Compiling all your code with these flags can't add them to these linked libraries. It's important to understand what you're linking against too. 

@david-a-wheeler suggested this might be worth a brief note in the Compiler Hardening Guide as it may be easily overlooked

thomasnyman:
Clarify compiler options hardening limitations when linking to pre-built artifacts

Fixes #705

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -116,6 +116,8 @@ Applications should work towards compiling warning-free. This takes time, but wa
 
 Compiler options hardening is not a silver bullet; it is not sufficient to rely solely on security features and functions to achieve secure software. Security is an emergent property of the entire system that relies on building and integrating all parts properly. However, if properly used, secure compiler options will complement existing processes, such as static and dynamic analysis, secure coding practices, negative test suites, profiling tools, and most importantly: security hygiene as a part of a solid design and architecture.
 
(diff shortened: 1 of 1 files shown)


jduck:
Looks good to me! Thanks!!

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -116,6 +116,8 @@ Applications should work towards compiling warning-free. This takes time, but wa
 
 Compiler options hardening is not a silver bullet; it is not sufficient to rely solely on security features and functions to achieve secure software. Security is an emergent property of the entire system that relies on building and integrating all parts properly. However, if properly used, secure compiler options will complement existing processes, such as static and dynamic analysis, secure coding practices, negative test suites, profiling tools, and most importantly: security hygiene as a part of a solid design and architecture.
 
+Hardened compiler options only take effect in code that is compiled with the hardened options. Consequently, compiler options hardening does not benefit software that has been pre-built before hardened options have been adopted. This is particularly a concern for projects that incorporate pre-built (possibly third-party) libraries or other components. In such cases, it is important to understand what components a project is being linked against, and how they in turn are built, to determine which components benefit from compiler options hardening.
```
david-a-wheeler, line 119: While code generation obviously only takes place during compilation, some compiler option flags are passed on to the linker and thus *can* have an impact after compilation. See in particular the `-Wl` options. So this needs a small weakening:

> s/Hardened compiler options only take effect/In most cases hardened compiler options only take effect/

> s/Consequently, compiler options hardening does not benefit software/Consequently, most compiler options hardening does not benefit software/

thomasnyman: Fixed in 844372a8ffb8f535e12a9abe73dcd879f17a5c8e.

david-a-wheeler (approved):

PR #716, opened 2024-12-23
Fix typo in return value from README (fixes #652)

### LINKED ISSUE #652: pySCG typo or error in CWE-703/CWE-392/compliant03.py
The [CWE-703/CWE-392/compliant03.py](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-392/compliant03.py)

code has an error on line 22
```py
        ret = list()
        invalid_arg = 0
        try:
            for res in result_gen:
                ret.append(res)
                invalid_arg += 1
            return ret
```
Not sure if its a typo, got flagged while using pyre.
Need to have a close look to understand the original intention of this code in order to fix it.

scovetta:
The code snippet incorrectly returns `res` when it should be returning `ret`. This doesn't really have an effect, since the sample demonstrates failing early.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-392/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-392/README.md
@@ -163,7 +163,7 @@ def map_threads(x):
             for res in result_gen:
                 ret.append(res)
                 invalid_arg += 1
-            return res
+            return ret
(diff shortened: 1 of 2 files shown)


s19110 (approved):
Yes, `ret` should be the value that is returned here. Thank you for catching this typo.

PR #732, opened 2025-01-29
Input2 rm yaml

david-a-wheeler:


david-a-wheeler:
Remove YAML from lab input2.

PR #782, opened 2025-02-05
Add copyright and license statements to each .js file

david-a-wheeler:
Make the license clearer per file.


david-a-wheeler:
This is already the license of checker.js and other materials, and the license is documented in the README, but making it clearer is a good plan.

PR #784, opened 2025-02-06
Add delay for a hint

david-a-wheeler:
Add delay for a hint. That way, instead of trying to use hints to write the answers, learners will be quietly encouraged to try to solve the problem themselves.

This adds a shorter delay (15 seconds) for each hint. If the user has seen a hint and hasn't changed anything, they can re-see the hint.

We earlier added a delay for "give up". This modifies the give up timer so that if they've asked for a hint, the give up timer starts from the time of the last hint. That way, people *can* give up, but they have to at least wait a little time.


david-a-wheeler:
I talked with CRob, who thought this tweak to our labs was reasonable. I'm going to merge this now, to discourage the overuse of lab hints while retaining full hint functionality. We can make further changes later (e.g., tweaking the hint delay time).

PR #795, opened 2025-02-20
Add PR review rules for translations

david-a-wheeler:


david-a-wheeler:
@gkunz @balteravishay (WG co-chairs) and @ninan27 @Muuhh-CTJ @shimos (proposed Japanese trusted translation team) - I propose this to ensure translations are reviewed. I think this proposal should be discussed at the next Best Practices WG meeting, and if agreed on, merged. Let me know if this is sensible.

balteravishay:
Sounds good to me. Thanks for the proposal!

david-a-wheeler:
[balteravishay](https://github.com/balteravishay) - you're welcome! I've added this PR to teh list of items to discuss at the next working group meeting: https://docs.google.com/document/d/1u1gJMtOz-P5Z71B-vKKigzTbIDIS-bUNgNIcfnW4r-k/edit?tab=t.0 - hopefully there will be general agreement & we'll move on.

The main issue at the moment is the Japanese translations of the labs. But we may as well set the general rule & then we know how to cover it.

docs/translations.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Translation review process
+
+To ensure that translations are good ones:
+
+1. All pull requests (PRs) that add or modify a translation
+   *must* be labeled as such.
+   The label is the language name, e.g., "Japanese" for Japanese.
+   Anyone can add that label, including the originator,
+   so that missing labels can be easily added.
```
gkunz, line 15: Can we verify that this is actually possible for any contributor? You probably need to be a member of the ossf org and/or have write permissions for this repo.

gkunz (commented):
I like this very much. Thank you @david-a-wheeler and particularly the translation team to push this forward. I just have one comment specifically on the management of labels.

docs/translations.md
david-a-wheeler: I meant "is allowed to by the process", not "GitHub permits this". If you're an originator, and you've added the PR in your own repo, I'm pretty sure you can add a label in your own repo.

david-a-wheeler: It's a fair point. I've tweaked the proposed text to clarify that case.

gkunz: Thanks, David. And in the meantime also I realized that I may have overcomplicated this in my head: with "adding a label", I was thinking of adding a new label to the list of labels, i.e., managing the labels not just adding an existing label to a PR. Anyhow, we absolutely don't need to over-complicate this. This updated version is ok for me.

david-a-wheeler:
I've created an example change so we can try out the process (as well as potentially fix a problem): https://github.com/ossf/wg-best-practices-os-developers/pull/798

I had to create the "Japanese" label, but now that it's created I think all should work well.

gkunz:
Thanks for creating the label - this step is what I had in mind when posting my prior comment. The test PR also looks good!

06kellyjac:
Is there interest in creating actual GH teams? Then the repo could leverage CODEOWNERS to auto request & require the team's review rather than this being a manual process/standard?

It looks like to make that useful translation teams would need repo write access which may be undesirable.

david-a-wheeler:
I added issue #801 to discuss using GitHub teams in the future. I thought it was more important to have *some* process now, and then we can refine it later.

PR #798, opened 2025-02-21
ja_input1.html: Fix button translations

david-a-wheeler:


david-a-wheeler:
This is a legitimate pull request, but it's *also* a test of the proposed translation review process as described in https://github.com/ossf/wg-best-practices-os-developers/pull/795

Do *NOT* merge this PR until a member of the Japanese trusted translation team has approved it. At the moment that would be one of:

* Non KAWANA - @ninan27
* Muuhh IKEDA - @Muuhh-CTJ
* Taku SHIMOSAWA - @shimos

shimos (approved):
@david-a-wheeler Thank you for the new review process and fixing the translation. The translation LGTM!

gkunz (approved):
Looks good to me and the new process seems to "work" 🙂

docs/labs/ja_input1.html
```diff
@@ -102,14 +102,14 @@ <h2>Interactive Lab (<span id="grade"></span>)</h2>
...
+<button type="button" class="hintButton">ヒント</button>
+<button type="button" class="resetButton">リセット</button>
+<button type="button" class="giveUpButton">諦める</button>
 <br><br>
 <p>
-<i>This lab was developed by David A. Wheeler at
+<i>このラボは、David A. Wheeler によって開発されました。
 <a href="https://www.linuxfoundation.org/"
->The Linux Foundation</a>.</i>
+> Linux Foundation</a>。</i>
```
06kellyjac, line 112: I'm still very much learning, especially with writing, so I apologize if this isn't useful :slightly_smiling_face:

With the current translation Linux Foundation just sits on it's own. Would this work better to convey the work was done at, for, or on behalf of the LF?

```html
<i>このラボは、<a href="https://www.linuxfoundation.org/">Linux Foundation</a> の David A. Wheeler によって開発されました。</i>
```

06kellyjac (commented):
Since this is already approved and matches the other document I wouldn't want to block merging the PR, but I thought I'd post a thought I had.

If this suggestion isn't accurate or helpful please let me know.

docs/labs/ja_input1.html
david-a-wheeler: @ninan27 @Muuhh-CTJ @shimos - would that be better? I defer to you as the experts on Japanese translations in the group.

The review process says that we *can* merge a PR once there's an approval, but we're not *obligated* to merge it instantly. It's quite reasonable to give some extra time so someone else can raise a question (like this).

shimos: @06kellyjac Thank you for pointing out!
The following would be slightly better because we basically do not use spaces for punctuations in Japanese:
```
<i>このラボは、<a href="https://www.linuxfoundation.org/">Linux Foundation</a>のDavid A. Wheelerによって開発されました。</i>
```

@david-a-wheeler To clarify, is the original English part ("at The Linux Foundation") means your affiliation? Otherwise, different translation might be more appropriate.

Muuhh-CTJ: According to "[Trademark Usage](https://www.linuxfoundation.org/legal/trademark-usage)" from The Linux Foundation;

> Use the full form of any trademarks (e.g., “The Linux Foundation”) in the first reference in all documents of mass communication, including marketing collateral and web pages.  You may then use any abbreviated or short form references (e.g. “TLF”) within the same the document where the full form has already been used.

I think "The Linux Foundation" is better than "Linux Foundation".  How about?

Article is a headache for us because Japanese has no word like articles.  I would like to hear the proper usage for the part from English speakers, sorry.

shimos: @Muuhh-CTJ (As you know, I'm not a native English speaker, but let me put some comments) I have checked several Japanese pages in The Linux Foundation. The interesting thing is that they are also using both "The Linux Foundation" and "Linux Foundation". Some pages like the official announcement, probably which originated from their English version, are using "The Linux Foundation." But some posts, which probably come from the Japan chapter of The Linux Foundation, are using "Linux Foundation" like [this](https://www.linuxfoundation.jp/press-release/2025/01/free-online-course-introduction-to-kubernetes-now-available-in-japanese/) .

Anyway, as it is the trademark, it'd be safer to use the official one even if it might sound a little awkward as Japanese.

```
<i>このラボは、<a href="https://www.linuxfoundation.org/">The Linux Foundation</a>のDavid A. Wheelerによって開発されました。</i>
```

06kellyjac: 提案を検討していただき、ありがとうございます。

Thank you for your input on the spacing, it's a bad habit of mine I'm working to address.

Using "The" with "Linux Foundation" in this sentence also makes sense. I can understand your caution as there are cases where it can be disjointed to use.

I'd read this back as either "This lab was created by The Linux Foundation's David A. Wheeler" or "This lab was created by David A. Wheeler of The Linux Foundation" which both work very well.

LGTM

david-a-wheeler: > To clarify, does the original English part ("at The Linux Foundation") mean your affiliation? Otherwise, different translation might be more appropriate.

Yes. My employer is The Linux Foundation.

It's unusual, but I believe the *official* answer is that its legal name in English _does_ start with the word "The". For example, if you look at its [legal bylaws](https://www.linuxfoundation.org/legal/bylaws) it says that the name of the organization is "The Linux Foundation". It's not the only one; I believe The Coca-Cola Company and The Walt Disney Company also begin with "The" in their official names.

I don't know how that is translated into Japanese. I'll also note that many people (even at the Linux Foundation) use a lowercase "the" or drop the "The" entirely in anything other than a formal legal document. So I guess I'll request "please do what appears the wisest thing to do in Japanese" because I truly don't know what the rules are regarding translation.

david-a-wheeler: > "This lab was created by The Linux Foundation's David A. Wheeler" or "This lab was created by David A. Wheeler of The Linux Foundation"

I agree, both of those English phrases seem completely correct to me, as I really am their employee.

shimos: @david-a-wheeler Thank you for your comment. Then, as in Japanese, we usually call it "Linux Foundation" in other than very rare official occasions,
```
<i>このラボは、<a href="https://www.linuxfoundation.org/">Linux Foundation</a>のDavid A. Wheelerによって開発されました。</i>
```
would be most appropriate.

Muuhh-CTJ: @shimos No objection.

Muuhh-CTJ (approved):

docs/labs/ja_input1.html
david-a-wheeler: @shimos @Muuhh-CTJ - thank you! I will use the translation recommended by @shimos that also received a "no objection" from @Muuhh-CTJ :

~~~~html
<i>このラボは、<a href="https://www.linuxfoundation.org/">Linux Foundation</a>のDavid A. Wheelerによって開発されました。</i>
~~~~

david-a-wheeler:
I made the updates I *think* were agreed to above. In both `ja_input1.html` and `ja_hello.html` we now have:

~~~~html
<i>このラボは、<a href="https://www.linuxfoundation.org/">Linux Foundation</a>のDavid A. Wheelerによって開発されました。</i>
~~~~

However, I'm a little confused by the discussion above. Is that what was agreed to?

Please approve this final version *or* propose an alternative. I want to ensure the correct version is merged.

Thank you!

shimos (approved):
@david-a-wheeler Thank you for the update. Yes, I think this one is what we agreed to.

06kellyjac (approved):
LGTM
Thanks everyone :smile:

david-a-wheeler:
Thank you *very* much, everyone! I realize these are small changes, but I hope they are helpful changes.

This pull request has also let us try our our new process of reviewing with members of our trusted translation team. We may improve the process later, but it's most important to *have* a process, we can improve it as we go.

Thank you!

PR #808, opened 2025-02-27
pySCG: main readme draft note links

myteron:
Adding links to the draft note to allow finding: 
* the web version when on GitHub 
* the GitHub version when viewing the web version


gkunz (approved):

myteron:
Hi @david-a-wheeler since we hope to have our own ossf repo at some stage, how persistent is https://best.openssf.org/Secure-Coding-Guide-for-Python/ as a URL?

david-a-wheeler:
It's *possible* have a separate repo, and then merge the results, but then you have to create extra steps to merge the results, extra steps to trigger things, etc. If there's no need to make things complicated, I suggest keeping things simple.

PR #812, opened 2025-03-07
fix: update sigstore cosign link

huang-julien:
Hellooo :wave: seems like this link leads to a 404


balteravishay (approved):
LGTM, thanks!

PR #821, opened 2025-03-20
Reorg front of Compiler Options Hardening Guide for clarity

david-a-wheeler:
Reorganize the front part of the
"Compiler Options Hardening Guide for C and C++"
for clarity.

There is a recent and very interesting post
"Mitigating a rsync Vulnerability: A Lesson in Compiler Hardening" by Mark Esler on March 19, 2025 at
<https://www.chainguard.dev/unchained/mitigating-a-rsync-vulnerability-a-lesson-in-compiler-hardening>

However, that post says "Chainguard implements most recommendations, and takes them a step further by also opting into..." and then adds options that our guidance *specifically* recommends. E.g., it adds `-fPIE` and `-PIC`, `-fsf-protection=full`, and so on.

This shows that even smart people who *read* the guidance think that only the list at the top is relevant, and don't even look at the table below it.

Let's put the table *immediately* after it, with bolded text to clarify what's meant. In addition, let's add `-Werror` to the table; the text below clarifies things, but if it's not even in the table, readers are likely to miss it.

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -17,7 +17,7 @@ This document focuses on recommended options for the GNU Compiler Collection (GC
 
 ## TL;DR: What compiler options should I use?
 
-When compiling C or C++ code on compilers such as GCC and clang, turn on these flags for detecting vulnerabilities at compile time and enable run-time protection mechanisms:
+When compiling C or C++ code on compilers such as GCC and clang, turn on these flags in all cases for detecting vulnerabilities at compile time and enable run-time protection mechanisms:
 
 ~~~sh
 -O2 -Wall -Wformat -Wformat=2 -Wconversion -Wimplicit-fallthrough \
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
@thomasnyman @gkunz - thoughts?

thomasnyman (approved):
lgtm, I had the same though while reading [Mark's blog post](https://www.chainguard.dev/unchained/mitigating-a-rsync-vulnerability-a-lesson-in-compiler-hardening) that it gives the impression the Compiler Hardening Guide only covers part of the options Chainguard uses, but  we concluded with Mark during [today's call](https://docs.google.com/document/d/1u1gJMtOz-P5Z71B-vKKigzTbIDIS-bUNgNIcfnW4r-k/edit?tab=t.0#bookmark=id.39t6zak24zfz) that the additions he proposes there were already covered.

david-a-wheeler:
Great! Mark's post is great, by the way, and it clearly shows that hardening flags can *really* harden software against attack. It's just that the option flags he specially added were... already in our guidance. I think this relatively small formatting tweak will help people better understand the guidance.

thomasnyman:
I agree. I'll go ahead and merge this.

PR #827, opened 2025-03-21
Release of The Memory Safety Continuum

balteravishay:
This PR adds the Memory Safety Continuum document to the best.openssf.org website.


david-a-wheeler (approved):

thomasnyman (approved):

david-a-wheeler:
@balteravishay - for now this seems like the right call.

In the longer term, do you want this document to look prettier, like the [Concise Guide for Developing More Secure Software
](https://best.openssf.org/Concise-Guide-for-Developing-More-Secure-Software)? If you do, we need to implement some nice way to generate the web pages. We typically use the [Simplest Possible Process (SPP)](https://best.openssf.org/spp/Simplest-Possible-Process), which uses GitHub Pages, unless you have a need for something fancier.

gkunz (approved):

PR #829, opened 2025-03-21
Clean up SPP documentation

david-a-wheeler:
Now that others are looking at applying the SPP,
it's clear we need to clean up its documentation.

This PR tries to do that. This has no technical changes, it's purely wording changes to make it easier to understand.


david-a-wheeler:
These are purely wordsmithing and formatting fixes, not technical changes, so I'll just merge these in.

PR #831, opened 2025-03-21
Add link from best.openssf.org to SPP

david-a-wheeler:


david-a-wheeler:
This will make the SPP documentation easier to find.

It has tips on how to edit the documents, so making it easy to find is a good thing.

gkunz (approved):

PR #832, opened 2025-03-22
docs: restructure concise evaluating guide,  add terminology list

janderssonse:
This restructure suggestion PR does the following:

The most important:

* Restructures the information (and does not remove any information from the original document) into simple tables with a theme header, and rules-description format. Aims to make the structure for the user clearer, and hit a add sweet spot between between, let's say a formal ISO 270x doc and something aimed at a developer/tech doing assessment.
* In the same mood, tightens up the language making it a bit more imperative, Verify that, Do that, ex.!

Other:
* Adds basic terminology list
* Adds a version table
* Adds a checkbox to that tables, (This is just an idea, and I don't know if I even like it myself, maybe not... let me know). Intended to use as a check-list, for the end user. Stay in, remove it?

This suggestion came from another PR where I asked about translations, and, while translating, I thought I would like to make the structure a bit more clearer in the original document.

### DIFF
--- a/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
+++ b/docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
@@ -1,57 +1,95 @@
 # Concise Guide for Evaluating Open Source Software
 
-_by the [Open Source Security Foundation (OpenSSF)](https://openssf.org) [Best Practices Working Group](https://best.openssf.org/), 2023-11-21_
+_by the [Open Source Security Foundation (OpenSSF)](https://openssf.org) [Best Practices Working Group](https://best.openssf.org/), 2025-03-28_
 
 As a software developer, before using open source software (OSS) dependencies or tools, identify candidates and evaluate the leading ones against your needs. To evaluate a potential OSS dependency for security and sustainability, consider these questions (all tools or services listed are merely examples):
 
-1. **Can you avoid adding it?** Can you use an existing (possibly indirect) dependency instead? Every new dependency increases the attack surface (a subversion of the new dependency, or its transitive dependencies, may subvert the system).
-2. **Are you evaluating the intended version?** Ensure you are evaluating the intended version of the software, not a personal fork nor an attacker-controlled fork. These techniques help to counter the common “typosquatting” attack (where an attacker creates an “almost-correct” name).
-   1. Check its name and the project website for the link.
-   2. Verify the fork relation on GitHub/GitLab.
-   3. Check if the project is affiliated with a foundation (in this case, you should be able to access the official source from the foundation’s website).
-   4. Check its creation time, and check its popularity.
-3. **Is it maintained?** Unmaintained software is a risk; most software needs continuous maintenance. If it’s unmaintained, it’s also likely to be insecure.
-   1. Has significant recent activity (e.g., commits) occurred within the last year?
-   2. When was its last release (was it less than a year ago)?
-   3. Is there more than one maintainer, ideally from different organizations?
-   4. Are there recent releases or announcements from its maintainer(s)?
-   5. Does its version string indicate instability (e.g., begin with “0”, include “alpha” or “beta”, etc.)
-4. **Is there evidence that its developers work to make it secure?**
-   1. Determine whether the project has earned (or is well on the way to) an [Open Source Security Foundation (OpenSSF) Best Practices badge](https://www.bestpractices.dev/).
-   2. Examine information on [https://deps.dev](https://deps.dev/), including its [OpenSSF Scorecards](https://github.com/ossf/scorecard) score and any known vulnerabilities.
-   3. Determine whether the package dependencies are (relatively) up to date.
-   4. Determine whether there is documentation explaining why it’s secure (aka an “assurance case”).
-   5. Are there automated tests included in its CI pipeline? What is its test coverage?
-   6. Does the project fix bugs (especially security bugs) in a timely manner? Do they release security fixes for older releases? Do they have an LTS (Long Time Support) version?
-   7. Do the developers use code hosting security features where applicable (e.g., if they’re on GitHub or GitLab, do they use branch protection)?
-   8. Identify security audits and whether any problems found were fixed. Security audits are relatively uncommon, but see OpenSSF’s “[Security Reviews](https://github.com/ossf/security-reviews)”.
-   9. Use [SAFECode’s guide _Principles for Software Assurance Assessment_](https://safecode.org/resource-managing-software-security/principles-of-software-assurance-assessment/) (2019), a multi-tiered approach for examining the software’s security.
-   10. Is the current version free of known important vulnerabilities (especially long-known ones)? Organizations may want to implement the [OpenChain](https://www.openchainproject.org/) [Security Assurance Specification 1.1](https://github.com/OpenChain-Project/Security-Assurance-Specification/tree/main/Security-Assurance-Specification/1.1/en) to systemically check for known vulnerabilities on ingest and as new vulnerabilities are publicly revealed.
-   11. Do they apply many practices in the [Concise Guide for Developing More Secure Software](https://best.openssf.org/Concise-Guide-for-Developing-More-Secure-Software)?
-5. **Is it easy to use securely?**
+## Initial Assessment
 
-   1. Are the default configuration and “simple examples” secure (e.g., encryption turned on by default in network protocols)? If not, avoid it.
-   2. Is its interface/API designed to be easy to use securely (e.g., if the interface implements a language, does it support parameterized queries)?
-   3. Is there guidance on how to use it securely?
+| Rule | Description | Done |
+|------|-------------|:--------:|
+| **Consider Necessity** | Evaluate whether the dependency can be avoided by utilizing existing components. Every new dependency increases the attack surface (a subversion of the new dependency, or its transitive dependencies, may subvert the system). |   |
+| **Verify Authenticity** | Verify that the software being evaluated is the authentic version from the authorized source, not a personal fork nor an attacker-controlled fork. These techniques help counter the common "typosquatting" attack (where an attacker creates an "almost-correct" name). Check its name and the project website for the link. Verify the fork relation on GitHub/GitLab. Check if the project is affiliated with a foundation. Check its creation time and popularity. |   |
 
-6. **Are there instructions on how to report vulnerabilities?** See the [Guide to implementing a coordinated vulnerability disclosure process for open source projects](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#guide-to-implementing-a-coordinated-vulnerability-disclosure-process-for-open-source-projects) for guidance to OSS projects.
-7. **Does it have significant use?** Software with many users or large users may be inappropriate for your use. However, widely used software is more likely to offer useful information on how to use it securely, and more people will care about its security. Check if a similar name is more popular - that could indicate a typosquatting attack.
-8. **What is the software’s license?** Licenses are technically not security, but licenses can have a significant impact on security and sustainability. Ensure every component has a license, that it’s a widely-used [OSI license](https://opensource.org/licenses) if it’s OSS, and that it’s consistent with your intended use. Projects that won’t provide clear license information are less likely to follow other good practices that lead to secure software.
-9. **What happens on a test addition?** Try adding the dependency as a test, preferably in an isolated environment, to investigate its impact:
-   1. Does it exhibit malicious behaviour, e.g., does it attempt to exfiltrate sensitive data?
-   2. Does it add unexpected or unnecessary indirect dependencies in production? For example, does it include production dependencies that are only required at development time or test time instead? If so, would their maintainers be willing to fix that? Every new dependency is a potential support problem or supply chain attack, so it's wise to eliminate unnecessary ones.
-10. **What are the results of code evaluation?** Even a brief review of software (by you, someone you hire, or someone else), along with recent changes to it, can give you some insight. Here are things to consider:
-    1. When reviewing its source code, is there evidence in the code that the developers were trying to develop secure software (such as rigorous input validation of untrusted input and the use of parameterized statements)?
-    2. Is there evidence of insecure/ incomplete software (e.g., many TODO statements)?
-    3. What are the “top” problems reported by static analysis tools?
-    4. Is there evidence that the software is malicious? Per [_Backstabber’s Knife Collection_](https://arxiv.org/abs/2005.09535), check the installation scripts/routines for maliciousness, check for data exfiltration from **~/.ssh** and environment variables, and look for encoded/ obfuscated values that are executed. Examine the most recent commits for suspicious code (an attacker may have added them recently).
-    5. Consider running the software in a sandbox to attempt to trigger and detect malicious code.
-    6. Consider running all defined test cases to ensure the software passes them.
-    7. See the [OpenSSF's list of security reviews](https://github.com/ossf/security-reviews/blob/main/Overview.md#readme).
+## Maintenance & Sustainability
 
-Other resources you may wish to consider include:
+Unmaintained software is a risk; most software needs continuous maintenance. If it's unmaintained, it's also likely to be insecure.
 
-1. [The Tidelift guide to choosing packages well (February 2021)](https://tidelift.com/subscription/choosing-open-source-packages-well), Tidelift
-2. [How to Evaluate Open Source Software / Free Software (OSS/FS) Programs](https://dwheeler.com/oss_fs_eval.html)
+| Rule | Description | Done |
+|------|-------------|:--------:|
+| **Activity Level** | Confirm significant recent activity (e.g., commits) has occurred within the previous 12 months. |   |
+| **Communication** | Verify the existence of recent releases or announcements from the project maintainer(s). |   |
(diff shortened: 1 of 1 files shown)


david-a-wheeler:
Definitely prettier! It's also _much_ longer (e.g., when printed). I don't know which is better for people. Does anyone else have any thoughts?

janderssonse:
Thanks. The terminology list could be removed as it takes up a lot of the new document. Maybe a terminology list could be an idea for a separate md-doc (in the future). But as for now, I could just remove it from the suggestion above.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+| **LTS (Long Term Support)** | A version of software that will receive maintenance and security updates for an extended period, often used for stability in production environments. |
+| **OpenChain Security Assurance Specification** | A specification from the [OpenChain Project](https://www.openchainproject.org/) (part of the Linux Foundation) that provides guidelines for managing security assurance in the open source supply chain. It helps organizations systematically identify and remediate known vulnerabilities in open source components. While OpenChain's license compliance framework became ISO/IEC 5230:2020, the [Security Assurance Specification](https://github.com/OpenChain-Project/Security-Assurance-Specification/) is a related initiative focusing specifically on security aspects. |
+| **OpenSSF Badge** | A certification from the [Open Source Security Foundation](https://openssf.org/) that indicates a project follows security best practices. Projects earn badges by meeting specific criteria for security, quality, and community engagement through the [Best Practices Badge Program](https://www.bestpractices.dev/). |
```
gkunz, line 101: To complement my suggestion above to include the new Security Baseline:

```suggestion
| **OpenSSF Badge** | A certification from the [Open Source Security Foundation](https://openssf.org/) that indicates a project follows security best practices. Projects earn badges by meeting specific criteria for security, quality, and community engagement through the [Best Practices Badge Program](https://www.bestpractices.dev/). |
| **OpenSSF Open Source Security Baseline** | A [collection of secure development requirements](https://baseline.openssf.org/) for open source projects relative to their maturity level. . |
```

gkunz (commented):
I in fact like the new layout, the check boxes and the languages changes. I think that the increase in length is outweighed by the more prominent structure. However, at the end of the day, it is a matter of taste.

Independent of the structure, I made two proposed additions to the content: the Security Baseline and API Stability as a new criterion. Probably we shouldn't mix structural changes and content changes, so I am happy to add them independently to the guide. Still, curious about feedback.

janderssonse:
Note: If you decide you want the checkbox to stay in, I will change it to '[ ]' instead of the current Unicode character □. The '[ ]' format is commonly used in markdown and makes it easy to tick off items, ie not just for printout. I was in a bit of a hurry when I added the Unicode character column - I should have used '[ ]' from the start.

agreedSkiing:
Colleague of @janderssonse chiming in. I would recommend to use "real" checkboxes `| <ul><li> [ ] </li></ul>  |` instead of `□` if they work in githubs markdown generator. Credit for this goes to https://stackoverflow.com/a/47344640

gkunz:
Thanks! As the guide is rendered to https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software, do we know what the output of `| <ul><li> [ ] </li></ul>  |` looks like there?

janderssonse:
Regarding Checkbox rendering:

I did some further experimentation and investigated both GitHub flavored markdown and the kramdown conversions (the rendering of the markdown in the link above is done with Jekyll, which uses kramdown GFM). I came to the conclusion that a Unicode character is the only solution that works for both cases I.E- rendering with GitHub's native renderer directly, and through Jekyll/Kramdown.

The official GitHub checkbox syntax - [ ] and above suggestion will render a checkbox in GitHub, but not in Jekyll/Kramdown.
Similarly, using an HTML input element would work for Jekyll/Kramdown, but gets removed from the GitHub native rendering as it sanitizes HTML.

**So, to summarize,  sadly, the best approach to support both cases (GitHub native rendering and Jekyll-kramdown-GFM rendering) is to use a Unicode character checkbox.**

Note:
I confirmed this through local conversion testing with: kramdown --input GFM withcheckboxvariants.md > a.html
(and AI would reach the same conclusion when provided with sufficient context).

gkunz:
Thanks @janderssonse for looking into this. I also confirmed the same result in the meantime.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+| **Parameterized Queries** | A technique for database access that separates SQL code from data, preventing SQL injection attacks. |
+| **Sandbox** | A controlled environment where untrusted code can be safely run without affecting the rest of the system. |
+| **Static Analysis** | Automated tools that examine code without executing it to identify potential bugs, vulnerabilities, or style issues. |
+| **Test Coverage** | A measure of how much of a codebase is executed during testing, indicating the thoroughness of your test suite. |
+| **Transitive Dependency** | Dependencies of your direct dependencies - packages your code relies on indirectly. |
+| **Typosquatting** | A malicious practice where attackers create packages with names very similar to popular packages, hoping developers will accidentally install them due to typos. |
+| **Vulnerability** | A weakness in software that could be exploited to cause harm or allow unauthorized access to systems or data. |
+**Document Version History**
```
maennchen, line 110:
```suggestion

**Document Version History**
```

Heading Is rendered as a table row:

image

janderssonse: Fixed the header as noted

janderssonse:
> We don't need the version number, the date is sufficient. We normally only use dates.

Fixed, versions removed, only using date

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+| **Best Practices Certification** | Determine whether the project has earned (or is well on the way to) an [Open Source Security Foundation (OpenSSF) Best Practices badge](https://www.bestpractices.dev/). | □ |
+| **Dependency Management** | Verify that the package dependencies are (relatively) up to date. | □ |
+| **Repository Security** | Confirm that developers use code hosting security features where applicable (e.g., if they're on GitHub or GitLab, do they use branch protection). | □ |
+| **Security Audits** | Identify existing security audits and verify that identified vulnerabilities have been fixed. Security audits are relatively uncommon, but see OpenSSF's "[Security Reviews](https://github.com/ossf/security-reviews)". | □ |
+| **Security Development** | Confirm they apply many practices in the [Concise Guide for Developing More Secure Software](https://best.openssf.org/Concise-Guide-for-Developing-More-Secure-Software). | □ |
```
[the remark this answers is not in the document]
janderssonse: Added as suggested.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+| **Security Scores** | Examine information on [https://deps.dev](https://deps.dev/), including its [OpenSSF Scorecards](https://github.com/ossf/scorecard) score and any known vulnerabilities. | □ |
+| **Testing Practices** | Evaluate if there are automated tests included in its CI pipeline and what is its test coverage. | □ |
+| **Vulnerability Status** | Confirm if the current version is free of known important vulnerabilities (especially long-known ones). Organizations may want to implement the [OpenChain](https://www.openchainproject.org/) [Security Assurance Specification 1.1](https://github.com/OpenChain-Project/Security-Assurance-Specification/tree/main/Security-Assurance-Specification/1.1/en) to systemically check for known vulnerabilities on ingest and as new vulnerabilities are publicly revealed. | □ |
+
+## Usability & Security
+
+| Rule | Description | Checkbox |
+|------|-------------|:--------:|
+| **Interface Design** | Verify that the interface/API is designed to be easy to use securely (e.g., if the interface implements a language, does it support parameterized queries). | □ |
```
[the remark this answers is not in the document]
janderssonse: Added as suggested.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+
+## Adoption & Licensing
+
+Licensing frameworks, while not directly security-related, significantly impact security and sustainability postures. Projects lacking clear licensing information frequently demonstrate deficiencies in other security best practices.
+
+| Rule | Description | Checkbox |
+|------|-------------|:--------:|
+| **License Clarity** | Verify that every component has a license, that it's a widely-used [OSI license](https://opensource.org/licenses) if it's OSS, and that it's consistent with your intended use. Projects that won't provide clear license information are less likely to follow other good practices that lead to secure software. | □ |
+| **Name Verification** | Check if a similar name is more popular - that could indicate a typosquatting attack. | □ |
+| **Usage Metrics** | Assess if it has significant use. Software with many users or large users may be inappropriate for your use. However, widely-used software is more likely to offer useful information on how to use it securely, and more people will care about its security. | □ |
```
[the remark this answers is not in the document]
janderssonse: Yes, I just used the original. I also think the sentence is not clear, and even that it should be clearer without that sentence. '...use. Widely-used....'

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
janderssonse: Added as suggested.

janderssonse:
I have rebased, and fixed most of the suggestions. Questions still:
- should checkbox stay or go?
- should terminology list stay or go?
- should the unclear sentence @gkunz mentions further up under "Usage Metrics" stay or be trimmed?

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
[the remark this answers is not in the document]
david-a-wheeler: I was thinking about the point of [You are not Google](https://blog.bradfieldcs.com/you-are-not-google-84912cf44afb): "if you’re using a technology that originated at a large company, but your use case is very different, it’s unlikely that you arrived there deliberately; no, it’s more likely you got there through a ritualistic belief that imitating the giants would bring the same riches."

Basically, while software with many users is often capable, it *might* not be a good match for your use case. You need to choose software relevant to your problem, not just choosing some software because it's the latest fad. The text is an attempt to counter fad-driven development aka [Hype Driven Development](https://blog.daftcode.pl/hype-driven-development-3469fc2e9b22).

I'm quite open to finding another way to say it.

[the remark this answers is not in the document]
david-a-wheeler: I think this should be a separate issue/PR. Let's not hold up the restructure for rewording this one point. This is the same wording as earlier.

david-a-wheeler:
My thoughts:

> should checkbox stay or go?

If we can get the checkboxes to "work" online they have some value. We use [Kramdown with Jekyll](https://jekyllrb.com/docs/configuration/markdown/) (the default for GitHub Pages). We can modify its configuration, but is there a configuration that makes them "work"? I'm not sure the Unicode checkboxes add much value, but it doesn't matter much to me either way.

> should terminology list stay or go?

I think it should stay. Having clear terms' definitions seems useful. Having them at the *end* means people can see the meat first.

> should the unclear sentence @gkunz mentions further up under "Usage Metrics" stay or be trimmed?

You mean "Software with many users or large users may be inappropriate for your use." yes? Let's make that a separate PR. I don't want to hold up this reorg while also trying to rewrite the text, and trying to do too many things at once can be confusing.

david-a-wheeler:
[janderssonse](https://github.com/janderssonse) found that Kramdown can't be configured to make this work. At first I thought we could configure Kramdown to make the checkboxes work ("that can't be right!") but I sure don't find a solution either :-(.

david-a-wheeler:
I found this which uses JavaScript to fix up some things. I'm not sure this effort is worth it: https://thatgurjot.com/til/clickable-checkboxes-on-jekyll-hugo/

david-a-wheeler:
In the long term we're working on an OpenSSF glossary. Once that has real content we could point off to that for the terminology.

janderssonse:
> [janderssonse](https://github.com/janderssonse) found that Kramdown can't be configured to make this work. At first I thought we could configure Kramdown to make the checkboxes work ("that can't be right!") but I sure don't find a solution either :-(.

As the checkbox currently seems to be of limited use, and is  causing problems for different renderings options..
I tried another idea - removing the checkbox character, leaving the table cell, with the header "Check" instead of checkbox. Still gives the impression to the reader that one should work through this guide and fill in the blanks. 

Next option would be that we just remove the Check-cell totally.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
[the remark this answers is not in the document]
janderssonse: Thanks for clarifying, I understand better now. While I understand and agree to the philosophical reasoning -  , I wonder if a reader would catch that, without a longer explanation. An idea would be to put the sentence under the  "Consider Necessity" question, and in a more straightforward form?  Something like -> Does your use case really motivate the addition, or are there simpler solutions? 

But, maybe the whole discussion as you suggest, should be moved to another issue.

janderssonse:
Rebased. Any chance of seeing this move forward? :)

What is left is, from what I can
- make a decision to leave an empty cell named check, which avoids any rendering problems for checkbox (done)
- make a decision about the "usage metrics" discussion earlier in the thread.

gkunz:
Hi @janderssonse. Sorry for the slow response times (Kubecon and other things happened). I support moving this forward and focusing this PR on restructuring the content. Specifically:

* I don't mind having no functioning checkboxes, but I think that adding a "check" column provides value by explicitly encouraging treating this as a checklist. I personally would have called the header "checked" instead of "check" (mainly because at first reading, I was expecting to find a check in this column) but this is very close to bike-shedding.
* let's move my feedback regarding some of the content to another PR - I'll take the action to propose new text once this PR here lands
* I also agree with keeping the glossary for now and move it later if needed.

gkunz (approved):
LGTM. I propose to move this forward.

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
```diff
@@ -1,57 +1,117 @@
...
+| **Security Reviews** | See the [OpenSSF's list of security reviews](https://github.com/ossf/security-reviews/blob/main/Overview.md#readme). |   |
+| **Static Analysis** | What are the "top" problems reported by static analysis tools? |   |
+| **Test Validation** | Consider running all defined test cases to ensure the software passes them. |   |
+
+## Additional Resources
+
+- [The Tidelift guide to choosing packages well (February 2021)](https://tidelift.com/subscription/choosing-open-source-packages-well), Tidelift
+- [How to Evaluate Open Source Software / Free Software (OSS/FS) Programs](https://dwheeler.com/oss_fs_eval.html)
+
+
```
david-a-wheeler, line 88:
```suggestion

```

You need to delete one of the lines to pass markdownlint.

janderssonse: Should be fixed now

janderssonse:
>  I notice that the markdowlinter complains so will fix that, and also change to your suggestion "Checked". I'll fix it later.

Both of them should now be fixed

docs/Concise-Guide-for-Evaluating-Open-Source-Software.md
david-a-wheeler: Great! Which blank line did you delete :-) :-) ?

david-a-wheeler:
To continue the bikeshedding, maybe `Checked` should say `Done?` - it's shorter and I think that's the intent.

janderssonse:
> To continue the bikeshedding, maybe `Checked` should say `Done?` - it's shorter and I think that's the intent.

Done... is Done!:)

gkunz:
Looks good to me. If we get another approving review, I'll merge this PR.

janderssonse:
> The new alphabetic organization seems to suggest that consulting [SAFECode’s guide _Principles for Software Assurance Assessment_](https://safecode.org/resource-managing-software-security/principles-of-software-assurance-assessment/) (2019) is the first thing to do. Originally this was point 9, as merely something you COULD do. Do we want to emphasize it that strongly?

How about changing "Utilize established... " to "Consider utililzing established ..." which should make it clear that you could, not should?

janderssonse:
> That helps, but s/utilizing/using/ ("utilizing" is an unnecessarily fancy word).

Rebased and utilized! Fixed I meant!

gkunz:
thanks @janderssonse. I'd like to really get this merged, so I propose doing so. The WG approved the restructure some time ago already.

Now, one last thing: there is a brand bew OpenSSF global glossary now: https://github.com/ossf/glossary

I suggest moving the terms from this doc to that glossary (I am also fine with first merging and then moving things around) @david-a-wheeler ?

janderssonse:
> thanks @janderssonse. I'd like to really get this merged, so I propose doing so. The WG approved the restructure some time ago already.
> 
> Now, one last thing: there is a brand bew OpenSSF global glossary now: https://github.com/ossf/glossary
> 
> I suggest moving the terms from this doc to that glossary (I am also fine with first merging and then moving things around) @david-a-wheeler ?

Thanks, I removed the glossary for now and will add to the global glossary in seperete PR, after this one is done.

david-a-wheeler:
Ok. I think we need merge, then improve further, or the bikeshedding will continue forever.

PR #863, opened 2025-04-24
📝 italic -> bold

ctcpip:
this changes most (but not all) existing italics to bold.  italics are not great for a11y, and suffers from particularly poor readability in the font that is used on [the website](https://best.openssf.org/Simplifying-Software-Component-Updates)

all emphasis remains -- just generally bold instead of italic


david-a-wheeler (approved):
This is purely a formatting change, not a content change, so let's try it out! We don't have a style guide that mandates one format or another. This was the point of soft-launching the content - it gives us a chance to make some last-minute changes, especially for formatting.

PR #868, opened 2025-04-28
pySCG: adding 838 as part of #531

myteron:
pySCG: adding CWE-838 as part of #531


docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-838/README.md
```diff
@@ -0,0 +1,102 @@
+# CWE-838: Inappropriate Encoding for Output Context
+
+Inappropriate handling of an encoding from untrusted sources or unexpected encoding can lead to unexpected values, data loss, or become the root cause of an attack.
+
+Mixed encoding can lead to unexpected results and become a root cause for attacks as showcased in [CWE-180: Incorrect behavior order: Validate before Canonicalize](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-180) and [CWE-175: Improper Handling of Mixed Encoding.](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-175/README.md) This rule showcases capturing the root cause by untrusted source its original binary without compromising the logging system for forensics.
```
s19110, line 5: I found this sentence a bit confusing while reading through the rule. How about something like this?
```suggestion
Mixed encoding can lead to unexpected results and become a root cause for attacks as showcased in [CWE-180: Incorrect behavior order: Validate before Canonicalize](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-180) and [CWE-175: Improper Handling of Mixed Encoding.](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-175/README.md) This rule showcases how to capture the root cause using an untrusted source's original binary without compromising the logging system for forensic purposes.
```

docs/Secure-Coding-Guide-for-Python/CWE-707/CWE-838/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```
+
+Trying to decode the modified encoded text in UTF-8 will result in the following exception:
+
+```bash
+UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 3: invalid start byte
+```
+
+Python is expected to use the `UTF-8` charset by default, which is backward compatible with `ASCII` [Python docs - unicode](https://docs.python.org/3/howto/unicode.html). Depending on the Python installation it may also be configured with any other encoding. It is recommended to always stick to `UTF-8` inside your program and do not bend the configuration of the OS [Batchelder 2022](https://www.youtube.com/watch?v=sgHbC6udIqc).
```
s19110, line 42: Punctuation + small rephrasing:
```suggestion
Python is expected to use the `UTF-8` charset by default, which is backward compatible with `ASCII` [Python docs - unicode](https://docs.python.org/3/howto/unicode.html). Depending on the Python installation, it may also be configured with any other encoding. It is recommended to always stick to `UTF-8` inside your program and not to bend the configuration of the OS [Batchelder 2022](https://www.youtube.com/watch?v=sgHbC6udIqc).
```

s19110 (approved):
Just a couple of rephrasing suggestions. Other than that, the rule looks good.

dwiley258:
No comments from me, rule looks good

PR #872, opened 2025-05-01
Correct a typo in oob1.html

Muuhh-CTJ:
Possibly a typo but please ignore if it is not.
This is a prep of my Japanese translation.


david-a-wheeler:
Oh my goodness, you're *exactly* correct! I've typed "OpenSSF" so often I didn't notice this mistake.

Thank you!

PR #873, opened 2025-05-01
Concise Evaluation Guide: Split recommendation in two separate ones

### LINKED ISSUE #853: Clarify inappropriate use text in concise guide for evaluating secure software
The concise guide to developing secure software says "Software with many users or large users may be inappropriate for your use." This doesn't seem to be clear enough, and there was a discussion here:
https://github.com/ossf/wg-best-practices-os-developers/pull/832#issuecomment-2788763536

Perhaps replace it with something like this:

> "Choose software that is a good solution for your problem, don't choose it merely because it's used by large companies or because it's the latest fad; avoid [Hype Driven Development](https://blog.daftcode.pl/hype-driven-development-3469fc2e9b22).

This is an issue, not a PR, because we're in the middle of reorganizing the concise guide & don't want to lose this idea.

gkunz:
Splitting the evaluation criterion about inappropriate suitability of a project into two seperate ones to make each one more clear:

- widely adopted projects have a higher change of being maintained,
- a user shall select a technology that solves the concrete problem at hand and not because it is the latest tech

Fixes #853


david-a-wheeler (approved):

david-a-wheeler:
Thank you, I think that's great! Simple and clear.

david-a-wheeler:
Splitting that one item into 2 separate ideas makes the concepts clearer. Thank you!

david-a-wheeler:
This fixes an issue, it has an approval, I see no objections, and it simply clarifies existing text. Let's merge it!

PR #874, opened 2025-05-01
Add a "consider" line or users

david-a-wheeler:
The text for component creators starts with "Consider".

This commit adds similar text to start the list for component users. I think this is important, as there are cases where you shouldn't blindly do something, but you *should* consider things. It also makes the document a little more consistent.


david-a-wheeler:
@gkunz ?

gkunz (approved):

PR #887, opened 2025-05-14
Japanese translation of regex1

Muuhh-CTJ:
Japanese translation for docs/labs/regex1.html and regex1.js


Muuhh-CTJ:
And appriciated if some mainter could label "Japanese" (and "Product: Labs")

shimos (approved):
@Muuhh-CTJ The translation LGTM!

PR #895, opened 2025-05-22
Fix the sample code in conversion.html

Muuhh-CTJ:
Issue : Sample code didn't call get_queue() and queue_count was not set.
Fix : Added the calling of get_queue() to set its return value for queue_count.

Trivial fix but a sample code should be consistent, I think.


david-a-wheeler:
I have no idea how this error slipped through. Thank you!

PR #896, opened 2025-05-22
Fix hint conditions of format-strings.js

Muuhh-CTJ:
The conditons should be intended as "user_format", instead of "user_input".


david-a-wheeler:
You're absolutely correct. Thank you!

PR #900, opened 2025-05-26
Japanese translation of format-strings

Muuhh-CTJ:
Japanese translaton for docs/labs/format-strings.html and format-strings.js
I appriciated if a mainter could label "Japanese" on this PR.


shimos (approved):
The translation LGTM

PR #906, opened 2025-05-29
Japanese translation of shell-injection

Muuhh-CTJ:
Japanese translaton for docs/labs/shell-injection.html and shell-injection.js
This PR should be labeled as "Japanese".


shimos (approved):
The translation LGTM!

PR #907, opened 2025-05-29
Japanese translation of sql-injection

Muuhh-CTJ:
Japanese translaton for docs/labs/sql-injection.html and sql-injection.js
@ninan27 @shimos , could you please review Japanese translation part?
This PR is also should be labeled as "Japanese".


shimos (approved):
Translation LGTM!

PR #915, opened 2025-06-11
Improve Japanese translation for labs

shimos:
This patch improves Japanese translation for some labs to be more coherent with the existing translation in the training course and easier to refer to the translated training course.

Please review the translation @Muuhh-CTJ  @ninan27


Muuhh-CTJ (approved):
Good points.  They are more consistent with LFD121-JP.
Approved.

PR #929, opened 2025-06-19
Remove mention of archived repo

andrewpollock:
This removes the mention of and outlink to the package-manager-best-practices repo, given the repo is archived.


gkunz:
Hi @andrewpollock. Just curious: Is this content being superseeded by work done in the security repos WG? If so, I am just  wondering if it would be useful to point to from the archived repo to that work.

andrewpollock:
@gkunz I can't say, sorry, I'm just trying to correct a deficiency that I'm seeing. Do you know who can speak to the archiving of the https://github.com/ossf/package-manager-best-practices repo and what the intended replacements for the content there are?

ctcpip (changes requested):
I don't think we should remove it just because it is archived.  It needs some work and should probably live in this repo, but if we think it still has useful/helpful information, we shouldn't erase the link to it.

PR #931, opened 2025-06-22
Add info on minimal UBSan runtime

yugr:
UBSan can be used in production environment if special minimal runtime is used.

PR #933, opened 2025-06-27
Add -fsanitize-minimal-runtime and -fsanitize-trap=undefined to UBSan description table

### LINKED ISSUE #326: Evaluate `-fsanitize-minimal-runtime` for possible hardening

thomasnyman:
Fixes #326 together with #931.


thesamesam:
See https://github.com/ossf/wg-best-practices-os-developers/issues/326#issuecomment-3014451880 for GCC.

thesamesam (approved):

gkunz (approved):
lgtm!

PR #934, opened 2025-06-28
Add additional considerations for -Wconversion

### LINKED ISSUE #819: The `-Wconversion` warning in the Compiler Options Hardening Guide should come with a caveat for brown-field projects
Raised during the discussion in the [2025-03-20 - C/C++ Compiler BP Guide](https://docs.google.com/document/d/1u1gJMtOz-P5Z71B-vKKigzTbIDIS-bUNgNIcfnW4r-k/edit?tab=t.0#bookmark=id.39t6zak24zfz) call.

The guide currently recommends enabling [implicit conversion warnings](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C%2B%2B#enable-implicit-conversion-warnings) using `-Wconversion` and `-Wsign-conversion`. The `-Wconversion` option, in particular, may be cumbersome for brown-field projects to enable as it tends to generate a large number of false positives due to the implicit conversions being very common in C. This is explicitly called out in the [GCC wiki page describing the `-Wconversion` feature](https://gcc.gnu.org/wiki/NewWconversion) also noting that _"[`-Wconversion`] is designed for a niche of uses [such as security audits] where the programmer is willing to accept and workaround invalid warnings."_ 

This type of caveat should be stated in a **When not to use?** or **Additional Considerations** section in the guide and encourage enabling `-Wconversion` for green-field projects.

Nevetheless, big projects such as [Chrome](https://issues.chromium.org/issues/40457839) have explored enabling `-Wconversion`. It would also be valuable to include practical advice for brown-field projects deploying `-Wconversion` such as hints suggested in:
- https://genodians.org/nfeske/2021-12-07-wconversion
- https://stackoverflow.com/questions/59654423/how-to-fix-errors-caused-by-wconversion
- https://issues.chromium.org/issues/40457839
- https://community.memfault.com/t/the-best-and-worst-gcc-compiler-flags-for-embedded-interrupt/90/15

Feel free to add your `-Wconversion` tips in the comments for this issue!

thomasnyman:
Fixes #819.

### DIFF
--- a/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
+++ b/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
@@ -289,8 +289,8 @@ In Clang, `-Wformat` includes the same diagnostics as `-Wformat=2`, but unlike i
 
 ### Enable implicit conversion warnings
 
-| Compiler Flag                                                                                             |       Supported since       | Description                         |
-|:--------------------------------------------------------------------------------------------------------- |:------------------------:|:----------------------------------- |
+| Compiler Flag                                                                                             | Supported since            | Description                         |
+|:--------------------------------------------------------------------------------------------------------- |:--------------------------:|:----------------------------------- |
 | <span id="-Wconversion">`-Wconversion`</span><br/><span id="-Wsign-conversion">`-Wsign-conversion`</span> | GCC 2.95.3<br/>Clang 4.0.0 | Enable implicit conversion warnings |
 
 #### Synopsis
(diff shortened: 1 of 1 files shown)


docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -310,6 +310,50 @@ If the resulting values are used in a context where they control memory accesses
...
+
+##### Replacing implicit casts with explicit ones
+
+Explicit casts are a temporary workaround, not a long-term solution. Each cast should eventually be:
+
+- Eliminated by harmonizing the types, or
+- Justified with a comment, or
+- Rewritten to reflect the programmer’s intent clearly (e.g., clamping values instead of truncating bits).
+
+Replacing implicit conversions with explicit C-style casts can introduce new bugs. For example, if an implicit cast from a 32-bit value to a 16-bit short is mistakenly replaced with a cast to an 8-bit char, this truncates more bits and misrepresents intent.
```
david-a-wheeler:
```suggestion
##### Removing implicit casts

Where practical, implicit casts that cause data loss should be refactored so that they are unnecessary. For example, they can be:

- Eliminated by harmonizing the types
- Rewritten to reflect the programmer’s intent clearly (e.g., clamping values instead of truncating bits).

If such a cast is necessary, convert it from an implicit cast to an explicit cast. Each such cast should be justified with a comment. A problem with replacing implicit conversions with explicit C-style casts is that they can introduce new bugs when used incorrectly. For example, if an implicit cast from a 32-bit value to a 16-bit short is mistakenly replaced with a cast to an 8-bit char, this truncates more bits and misrepresents intent.
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -310,6 +310,50 @@ If the resulting values are used in a context where they control memory accesses
 
 For C++ warnings about conversions between signed and unsigned integers are disabled by default unless `-Wsign-conversion` is explicitly enabled.
 
+#### Additional Considerations
```
david-a-wheeler, line 320:
```suggestion

This warning only applies where a data conversion may lead to data loss. C and C++ conditionals (such as after an "if" statement) accept types other than booleans, for example, integers (where non-zero is interpreted as true) and pointers (where non-NULL is interpreted as true). This interpretation of truthiness could be considered a data type conversion (in a sense), but this option does not create spurious warnings about these perfectly reasonable constructs.

#### Additional Considerations
```

gkunz (approved):
LGTM

Issue #201, opened 2023-07-11
Compiler attributes for improved security


siddhesh:
A number of compiler attributes (e.g. `alloc_size`, `malloc`, `access`, `nonnull`) provide additional information to the compiler to allow it to (1) detect potentially undefined behaviour and (2) optimize code. These often end up in the win-win category where the compiler is able to improve performance as well as security.  I had documented a [couple of them in an article](https://developers.redhat.com/articles/2023/02/06/how-improve-application-security-using-fortifysource3#) in the past in a narrow scope, but there's scope for a more comprehensive document.  There are some caveats to their usage though, so it's likely not a matter of simply listing them.

Would a new document describing these attributes and how OSS developers could use them with best effect be a useful addition to this project?

david-a-wheeler:
The Linux kernel has set up a set of conventions to prevent buffer overwrites in many circumstances, that might be a useful source.

We currently have a project for compiler options, but that's a different thing.

The issue would be that some people would need to dedicate real time for this. If that can be done, I imagine it could be really helpful.

siddhesh:
I don't want to promise (because this is largely going to be free time work for me and it's summer!) but I can take a stab at it at some point. I just wanted to put this issue out here to (1) get a sense for whether this would be interesting for this WG and (2) remind myself of it later whenever I have some free time to tackle it.

thomasnyman:
Fully supporting this! A new document seems to be a sensible approach in my opinion, although there's potentially some overlap with the existing guide too, e.g. [this recent RFC](https://discourse.llvm.org/t/rfc-enforcing-bounds-safety-in-c-fbounds-safety/70854) for Clang proposes a combination of option flag + annotations.

david-a-wheeler:
@siddhesh - could you take a quick stab at a starting document? I think there's general agreement that this is useful, but probably needs to be a separate document. The big difference, as I understand it, is that this involves changes to the source code itself. Feel free to focus on just C (or just C++); I suspect that it may need to be more specific about the language.

I encourage looking at:

* Some of the approaches and discussion of the Linux kernel (current and proposed) 
* GCC C extensions: https://gcc.gnu.org/onlinedocs/gcc/C-Extensions.html

siddhesh:
[...]

PR #937, opened 2025-07-01
Update wording on `__attribute__((malloc))`

Quuxplusone:
- Clang doesn't support the two-argument version. This is https://github.com/llvm/llvm-project/issues/53152 .
- Clang 21 isn't a thing yet.
- Emphasize the "optimization" we're talking about (noalias); I believe that's the only optimization we're talking about here (and if not, we should be even more explicit, because it's fooled me). Add an example of the optimization to the relevant Godbolt, since it was missing any example before.
- Fix two trivial "cplusplis" typos.

Addresses my comments in #201.


docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -75,7 +75,8 @@ In GCC, the `malloc (`_`deallocator`_`)` and `malloc (`_`deallocator`_`,` _`ptr-
 - **Memory leaks** (`-Wanalyzer-malloc-leak`) if if there is an execution path in which the result of an allocation call goes out of scope without being passed to the deallocation function.
 - **Invalid free** (`-Wanalyzer-free-of-non-heap`) if a deallocation function is used on a global or on-stack variable.
 
-Clang supports both forms of the `malloc` attribute but does not yet implement the `-Wmismatched-dealloc` and `-Wmismatched-new-delete` warnings. Instead, Clang provides the `ownership_returns`, `ownership_takes`, and `ownership_holds` attributes[^clang-ownership]: that interact with the Clang static analyzer[^clang-checkers].
+Clang's parser recognizes both forms of the `malloc` attribute, but Clang does not yet implement the `-Wuse-after-free`, `-Wmismatched-dealloc`, and `-Wmismatched-new-delete` warnings
+that would use that information. Instead, Clang provides the `ownership_returns`, `ownership_takes`, and `ownership_holds` attributes[^clang-ownership]: that interact with the Clang static analyzer[^clang-checkers].
```
thomasnyman: Nit: superflous linebreak

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -101,12 +102,16 @@ void my_free(void *ptr);
 void *my_malloc(size_t size) __attribute__ ((malloc, malloc (my_free, 1)));
 ~~~
 
-Note that to benefit both from the associated optimizations and improved detection of memory errors functions should be marked with _both_ the form of the attribute without arguments and the form of the attribute with one or two arguments. [[Extended example at Compiler Explorer](https://godbolt.org/z/bc97ahbnd)]
+Note that to benefit from both the associated `noalias` optimization and the improved detection of memory errors,
+functions should be marked with _both_ the zero-argument `malloc` attribute and the form of the attribute with one
```
thomasnyman: Nit: superflous linebreaks likely due to editor line wrap settings. To be consistent with the rest of the guide it'd be better to format the paragraph as a single line.

thomasnyman (changes requested):

thomasnyman:
Thanks for the contribution. Content-wise I think this looks good. Left a couple of nits on the formatting to make it consistent with the rest of the guide. In addition, for us to consider these changes, you would need to to confirm the [LF DCO](https://developercertificate.org/) by leaving your sign-off in the commit message since the OpenSSF Charter requires DCOs for all inbound contributions (see https://github.com/ossf/tac/blob/main/dco.md).

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -101,12 +102,16 @@ void my_free(void *ptr);
 void *my_malloc(size_t size) __attribute__ ((malloc, malloc (my_free, 1)));
 ~~~
 
-Note that to benefit both from the associated optimizations and improved detection of memory errors functions should be marked with _both_ the form of the attribute without arguments and the form of the attribute with one or two arguments. [[Extended example at Compiler Explorer](https://godbolt.org/z/bc97ahbnd)]
+Note that to benefit from both the associated `noalias` optimization and the improved detection of memory errors,
```
thesamesam: I don't see why we're talking about `noalias` here, which is an LLVM IR term. Nor is it really useful I think to talk about specifics like that here, it's prone to becoming stale and isn't going to apply to GCC.

I also think it's kind of obvious that there'll be optimisations from something knowing is a `malloc` and `noalias` doesn't particularly tell us much.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
```diff
@@ -56,7 +56,7 @@ Attributes influence not only diagnostics generated by the compiler but also the
...
 |:--------------------------------------------------------------------------------------------|:---------------------------:|:----------------------------:|:------------------------------------------------------------------------------------------------- |
 | <span id="malloc">`malloc`</span>                                                           | GCC 2.95.3<br/>Clang 13.0.0 | Function                     | Mark custom allocation functions that return non-aliased (possibly NULL) pointers.                |
-| <span id="malloc (dealloc)">`malloc (`_`deallocator`_`)`</span>                             | GCC 11.0.0<br/>Clang 21.0.0 | Function                     | Associates _`deallocator`_ as the valid deallocator for the storage allocated by marked function. |
+| <span id="malloc (dealloc)">`malloc (`_`deallocator`_`)`</span>                             | GCC 11.0.0                  | Function                     | Associates _`deallocator`_ as the valid deallocator for the storage allocated by marked function. |
```
thesamesam, line 59: I think not having Clang here at all is confusing as it implies you can't safely use the attribute, no? You can safely use it, just it doesn't do anything.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
Quuxplusone: Changed "associated `noalias` optimization" to "no-alias optimization." I think it's important to

> Emphasize the "optimization" we're talking about (noalias); I believe that's the only optimization we're talking about here (and if not, we should be even more explicit, because it's fooled me).

I don't think it's important to shy away from the term "noalias," since that is literally the optimization we're talking about: the optimization that says "This thing that by default you think alias? It no alias." 😛 The LLVM IR terminology comes from the nature of the optimization, not the other way around (and actually — although this surprised me too — when you google "noalias optimization," the first two hits are in Visual Studio docs, not LLVM docs!).

I can't parse your second sentence: "I also think it's kind of obvious that there'll be optimisations from something knowing is a `malloc` and `noalias` doesn't particularly tell us much."

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
Quuxplusone: It produces a warning (or an error with `-Werror`), which in practice _should_ break your build. If you're not compiling with `-Werror` — or if you're letting warnings happen and not fixing them — I think that's _vastly_ more of a departure from "security best practices" than omitting to use some attribute.

I guess one criterion to answer "Does Clang support this attribute?" is to see whether a cross-platform build system like CMake or Meson believes that Clang supports it. I grepped github for `__attribute__((malloc,` and one of the top hits was this commit to Meson: https://github.com/OpenRC/openrc/commit/17de4e5dfdf614332091b856bc05837b2e204a5a — 
```
# Meson's has_function_attribute doesn't know about GCC's extended
# version (with arguments), so we have to build a test program instead.
malloc_attribute_test = '''#include<stdlib.h>
__attribute__ ((malloc (free, 1)))
int func() { return 0; }
'''
if cc.compiles(malloc_attribute_test, name : 'malloc attribute with arguments')
  add_project_arguments('-DHAVE_MALLOC_EXTENDED_ATTRIBUTE', language: 'c')
endif
```

When you feed this to Meson on a Clang system, does it indeed set `-DHAVE_MALLOC_EXTENDED_ATTRIBUTE=1`? If it does, then I'll (grudgingly :)) accept that _in practice_ Clang trunk _is regarded as_ supporting the extended malloc attribute, even though [the Clang maintainers themselves say it's not supported yet](https://github.com/llvm/llvm-project/issues/53152). Sadly I don't have a Meson system to test this on myself.

thomasnyman: I can understand both viewpoints on this. I also see that listing Clang under the "Supported since" column give the wrong impression especially given the current behavior that issues a warning. What I might suggest is to add a note Clang in the description to the effect of "Clang trunk has this attribute, but lacks the corresponding diagnostic (resulting in -Wignored-attributes warning on use)." which we'll update to say Clang 21.0.0 once it is released.

thesamesam: > was this commit to Meson

Right, my commit to a project using Meson (not Meson itself).

> Sadly I don't have a Meson system to test this on myself.

I think we've maybe reached an impasse as I don't have a Clang trunk system to test it on ;)

If I extract the command line used and the program from `meson-log.txt` on [godbolt](https://godbolt.org/z/3v9oWrPsc), I see this:
```
<source>:2:17: warning: 'malloc' attribute only applies to return values that are pointers [-Wignored-attributes]
    2 | __attribute__ ((malloc (free, 1)))
      |                 ^
    3 | int func() { return 0; }
      | ~~~
1 warning generated.
Compiler returned: 0
```

so I think yes, it'll pass?

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
thesamesam: I think `noalias` is LLVM-specific terminology and I don't see a benefit to using it.

> I can't parse your second sentence: "I also think it's kind of obvious that there'll be optimisations from something knowing is a malloc and noalias doesn't particularly tell us much."

I'm saying that it's either:
a) it's LLVM-specific terminology and you're using it to aid researching what LLVM does, or
b) it's obvious that the `malloc` attribute is going to help with that optimisation, so `noalias` isn't telling you anything

I really don't see a point in adding it. We could add a lot of details like this throughout the guide for e.g. the `access` attribute and GCC's object size pass, I don't see how it helps the reader. The guide is already quite verbose.

If you would really like to include the information, please explain it in detail instead of `noalias`, but I'd still say that's unnecessary and irrelevant for this document.

docs/Compiler-Hardening-Guides/Compiler-Annotations-for-C-and-C++.md
eli-schwartz: > It produces a warning (or an error with `-Werror`), which in practice _should_ break your build. If you're not compiling with `-Werror` — or if you're letting warnings happen and not fixing them — I think that's _vastly_ more of a departure from "security best practices" than omitting to use some attribute.

The idea that -Werror must be used and fail the build on any form of error is an admirable one, but it necessarily has a rough interaction with the fact that C/C++ compilers don't have a structured way to distinguish between coding style (opinionated) warnings and semantic warnings.

In general, it is most emphatically NOT a code smell or a violation of security best practices to react to warnings -- or even -Werrors -- by determining the warning is too much of "opinionated lint" an silencing it via `-Wno-ignored-attributes` or similar.

So, eminently reasonable to safely use it everywhere?

Quuxplusone: Meh. I think OSSF and I just have different opinions about the meanings of the words "safe" and "best practices." Like, if a feature is not supported on one of my targeted compilers, I'd use a macro to avoid it; e.g.

```
#if MY_COMPILER_GCC
 #define MY_ATTR_MALLOC __attribute__((malloc, malloc(free)))
#else
 #define MY_ATTR_MALLOC
#endif

[...]

int *func() MY_ATTR_MALLOC { return malloc(sizeof(int)); }
```

Then I'd "treat warnings as errors" and make sure my code compiled cleanly on all of my targeted compilers.

It sounds like the OSSF way is to put the attribute unconditionally, and then use `-Wno-ignored-attributes` in the project file (because by definition the project cannot be made clean w.r.t. that warning anymore). In my style, that would be frowned on; we'd prefer to fix the warning rather than disable it for the whole team.

Anyway, as I'm not part of OSSF, I'll abandon this PR.

PR #940, opened 2025-07-10
CWE-1095: show that list modification is dangerous

hedrok:
Before this commit the wording was that modifying list works but is not recommended.

But it works as long as no two consecutive elements are deleted, otherwise part of elements is not checked at all without any exceptions raised.

Changed README.md, compliant01.py and noncompliant01.py to demonstrate that.


BartKaras1128 (approved):
Thanks for the PR! I think this example is a more interesting way of showing that modification of a list in a loop can be dangerous and could have unexpected outcomes.
Approved by me.

myteron (approved):
Thank you @hedrok for the contribution. Much better example.

hedrok:
Thanks! :)

PR #946, opened 2025-07-17
CWE-476 

BartKaras1128:


docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-476/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+print(name)
+
+```
+
+The `is_valid_name()` function could get called with a null argument, which raises an `AttributeError` due to the `NoneType` object not having any `split` attribute.
+
+## Compliant Code Example - Verify that the object is not None before accessing or modifying its fields
+
+The `compliant solution` includes the same implementation as the previous `non-compliant example` but now checks if the string is `None`.
```
s19110: I think some backticks are unnecessary here:
```suggestion
The `compliant solution` includes the same implementation as the previous non-compliant example but now checks if the string is `None`.
```
```suggestion
The compliant solution includes the same implementation as the previous `non-compliant example` but now checks if the string is `None`.
```

docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-476/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+#####################
+# Attempting to exploit above code example
+#####################
+print_number_of_students(None)
+
+```
+
+The code example attempts to directly calling `len()` on a non-array value, but rather on `None`, which will raise a `TypeError`. Such unchecked dereferencing of a `None` value could result in an application crash or denial of service, impacting system availability.
```
s19110: This sentence sounds a bit odd. How about something like that?
```suggestion
The code example attempts to directly call `len()` on a non-array value, in this case on `None`, which will raise a `TypeError`. Such unchecked dereferencing of a `None` value could result in an application crash or denial of service, impacting system availability.
```

s19110 (changes requested):
I left a couple small cosmetic suggestions. The rule looks good. Remember to also add the README link to the table in the main readme of the Secure Coding project.

andrew-costello:
The rule looks good to me, I agree with @s19110 's rewording suggestion, I was going to put it in myself. When it is reworded, I will review again and give +1.

docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-476/README.md
myteron: not sure if I would phrase it like this "The `compliant01.py` code includes the same implementation as the previous `noncompliant01.py` " would be easier to just say that compliant01.py is checking for None. Anyways, aggreed to only apply cosmetics by naming the file name's and putting them into backticks.

PR #953, opened 2025-07-31
Add Data Protection to Security-Focused Guide for AI Code Assistant Guide

rozhukov:
Hey folks, I realized I might be late to the party with my edits in the doc a few weeks ago https://docs.google.com/document/d/16yydtZ6vFAJhF1i_0uPRY7nLN3wtjxU7aq_WIuF2YKs/edit?tab=t.0#heading=h.7clqkb6tjf05 thus creating this PR directly here to suggest adding Data Protection bullet point explicitly (even though it's partly mentioned in logging section above) as it's often overlooked as a good design pattern. I also think it will not unreasonably expand this concise guidance, which I very much like BTW.


balteravishay:
@rozhukov, please check out the main branch again, it was added by me after you've suggested that to the google docs sheet. thanks for your contribution!

PR #957, opened 2025-08-08
Specifically note constant-time comparisons and auth flow

david-a-wheeler:
Extend the AI guidance to specifically discuss constant-time comparisons. It's common to fail to do this, so
give specific instructions to counter this tendency.


david-a-wheeler:
@balteravishay - thoughts?

balteravishay:
looks great, thanks @david-a-wheeler !

balteravishay (approved):

PR #958, opened 2025-08-08
Add a few tweaks to AI guidance

david-a-wheeler:
You might want to add some of this to specific prompts instead of the custom general instructions, so make it clear that's a valid use.

Discuss more about asking the AI to do analysis. While they can do analysis, I've found through experimentation that they are FAR more effective if the human provides specific relevant facts as part of the request for analysis. It's really extraordinary when humans provide expert guidance - the AI assistant sometimes jumps up from "not smart assistant" to "brilliant careful analysis" when it's the human provides it non-obvious expert information.


balteravishay (approved):
LGTM, Thanks!

PR #963, opened 2025-08-13
Add more about asking LLM to be an expert from Dilgren

david-a-wheeler:


balteravishay (approved):
great addition, thanks!

PR #964, opened 2025-08-14
pySCG: adding CWE-397: Declaration of Throws for Generic Exception

s19110:


dwiley258 (commented):
Looks good to me

docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-397/README.md
```diff
@@ -0,0 +1,180 @@
+# CWE-397: Declaration of Throws for Generic Exception
+
+Avoid raising exceptions that aren't informative on specific errors.
+
+`Exception` is the base class of all non-system-exiting exceptions [[Python docs 2025](https://docs.python.org/3/library/exceptions.html#Exception)]. Specific exceptions allow to determine why they were raised and attempt recovery. To catch a generic `Exception`, all more specific exceptions that subclass it would have to be caught. Raising `Exception` is likely to hide bugs and and prevents using more specialized except statements. The problem is even more severe when raising `BaseException`, which additionally includes subclasses signifying termination signals, such as `KeyboardInterrupt` and `SystemExit` [[PEP 2024](https://peps.python.org/pep-0352/#exception-hierarchy-changes)].
```
BartKaras1128: Found this a bit hard to read. I suggest rewording it in the following way:

```suggestion
`Exception` is the base class of all non-system-exiting exceptions [[Python docs 2025](https://docs.python.org/3/library/exceptions.html#Exception)]. Specific exceptions allow to determine why they were raised and attempt recovery. To catch a generic `Exception`, you must also account for any more specific exceptions that inherit from it. Raising `Exception` is likely to hide bugs and and prevents using more specialized except statements. The problem is even more severe when raising `BaseException`, which additionally includes subclasses signifying termination signals, such as `KeyboardInterrupt` and `SystemExit` [[PEP 2024](https://peps.python.org/pep-0352/#exception-hierarchy-changes)].
```

BartKaras1128 (approved):
Left one small comment but I approve otherwise!

gkunz (approved):

PR #967, opened 2025-08-26
new branch to fix issues when rebaseing

dwiley258:


docs/Secure-Coding-Guide-for-Python/CWE-703/CWE-230/README.md
```diff
@@ -0,0 +1,150 @@
+# CWE-230: Improper Handling of Missing Values
+
+The `NaN` value should be stripped before as they can cause surprising or undefined behaviours in the statistics functions that sort or count occurrences [[2024 doc.python.org]](https://docs.python.org/3/library/statistics.html).
```
BartKaras1128, line 3:
```suggestion
The `NaN` value should be stripped before as they can cause surprising or undefined behaviours in the statistics functions that sort or count occurrences [[2024 doc.python.org]](https://docs.python.org/3/library/statistics.html).

```

BartKaras1128 (approved):
Just think maybe the first sentence should be it's own paragraph, for most rulesets we have the first sentence being a sort of "do this" or "don't do this" kind of rule. Or else maybe you could have a whole other sentence if you feel like it? I approve it otherwise, looks good to me

PR #973, opened 2025-09-04
adding 460 to main table, fixing cosmetics and formatting

myteron:
460 was missing on main readme.md for pySCG


docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-460/README.md
```diff
@@ -10,92 +10,102 @@ A consequence of this is that the code is left in a bad state.
...
+    lock.release()
+
+```
+
+It is best practice to use `with` statement in such cases as it will make sure the resource gets released even if an exception occurs in the execution.
 
 ## Non-Compliant Code Example
 
-In the noncompliant.py example, a thread gets locked, but not unlocked due to an exception being thrown before it can be closed. This might lead to the lock remaining closed and inaccessible for further use.
+In the `noncompliant.py` example, a thread gets locked, but not unlocked due to an exception being thrown before it can be closed. This might lead to the lock remaining closed and inaccessible for further use.
```
s19110: Because we number code examples even when there is only one of each, and there is no `noncompliant.py` file in the project, the name in the text should correspond to the name of the file:
```suggestion
In the `noncompliant01.py` example, a thread gets locked, but not unlocked due to an exception being thrown before it can be closed. This might lead to the lock remaining closed and inaccessible for further use.
```

docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-460/README.md
```diff
@@ -10,92 +10,102 @@ A consequence of this is that the code is left in a bad state.
...
+
+# Next attempt to acquire the lock will block forever; as there is a deadlock!
 lock.acquire()
-print("This will never print because the lock was never released.")
+print("This will not print because the lock was never released.")
 
 ```
 
-In the above code example, the acquired lock never gets released, as an error gets thrown before it can be released.
+In the `noncompliant.py` code example, the acquired lock never gets released, as an error gets thrown before it can be released.
```
s19110:
```suggestion
In the `noncompliant01.py` code example, the acquired lock never gets released, as an error gets thrown before it can be released.
```

docs/Secure-Coding-Guide-for-Python/CWE-664/CWE-460/README.md
```diff
@@ -10,92 +10,102 @@ A consequence of this is that the code is left in a bad state.
...
-def compliant_example():
+
+def perform_critical_operation():
     with lock:
         # the lock has been acquired using the 'with' statement and will be released when the block exits; even if an exception occurs
         print("Lock acquired, performing critical operation...")
         # raising an exception
         raise ValueError("Something went wrong!")
+        # This line will not be reached because of the exception above,
     print("Lock released.")
```
s19110: Removing a comma at the end of the comment. Also, there wasn't a comment for raising an exception in `noncomplian01.py` and since that part doesn't change in this code example, I think we can remove that comment:
```suggestion
def perform_critical_operation():
    with lock:
        # the lock has been acquired using the 'with' statement and will be released when the block exits; even if an exception occurs
        print("Lock acquired, performing critical operation...")
        raise ValueError("Something went wrong!")
        # This line will not be reached because of the exception above
    print("Lock released.")

```

Same change applies to compliant01.py

s19110 (approved):

PR #993, opened 2025-09-19
pySCG: adding 783 as part of #531

myteron:
part of #531

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
@@ -1,0 +1,176 @@
+# CWE-783: Operator Precedence Logic Error
+
+Failing to understand the order of precedence in expressions that read and write to the same object can lead to unintended side effects.
+
+Python has distinct different concepts for:
+
+<table>
+<tr>
+<th>
+type
+</th>
+<th>examples</th>
+<th>Typical direction</th>
+<tr>
+<td>Assignments</td><td>to store a value such as x = 1</td><td>right-to-left.</td>
+</tr>
+<tr>
+<td>Expressions</td><td>3+4 or x*2</td><td>left-to-right</td>
+</tr>
+<tr>
+<td>Augmented assignments</td><td>a += 1</td><td>left-to-right <a href="https://docs.python.org/3/reference/simple_stmts.html?highlight=augmented%20assignment%20operators#augmented-assignment-statements">[Python docs 2025 - simple statements]</a></td>
+</tr>
+</table>
+
+Expressions such as `2 ** 3`, or two to the power of three, are evaluated from right to left [[python power 2025](https://docs.python.org/3/reference/expressions.html#index-59)] as demonstrated in `example01.py`
+
+_[example01.py:](example01.py)_
+
+```py
+"""Code Example"""
+
+print(2**3**2)  # prints 512
+print((2**3) ** 2)  # prints 64
+print(2**9)
+
+```
+
+The first expression would print `64` if Python would resolve from left-to-right but prints `512` as it calculates `3**2` before using its result with `2**9`.
+The `example02.py` behaves 'normal' for a programmer but makes no  sense as a mathematical formular.
+
+_[example02.py:](example02.py)_
+
+```py
+z = 2
+z *= 2 + 1
+print(f"z *= 2 + 1    ={z}")
+```
+
+If a method changes an object’s state (has side effects) and is called multiple times within one expression, the result can be surprising and incorrect. For further info on python's order of precedence refer to The Python Language Specification , §6.16, "Evaluation Order" [[PLR 2022](https://docs.python.org/3/reference/expressions.html#evaluation-order)].
+
+## Non-Compliant Code Example
+
+This `noncompliant01.py` code is expected to provide labels for numbers obfuscates the evaluation and logic.
+
+_[noncompliant01.py](noncompliant01.py):_
+
+```python
+"""Non-Compliant Code Example"""
+
+
+def label(number: int) -> str:
+    a = int(number < 0)  # negative flag
+    b = (number & 1) ^ 1  # even flag (1 for even, 0 for odd)
+    c = int(number < 5)  # small flag
+
+    key = (a << 2) | (b << 1) | c  # pack flags into a single key
+
+    parts = ("big", "small", "even", "even", "neg", "neg", "neg", "neg")
+
+    permuted = tuple(parts[(i * 5) & 7] for i in range(8))
+
+    idx = (key * 5) & 7
+    return permuted[idx]
+
+
+for number in range(-3, 3):
+    print(f"{number} = {label(number)}")
+
+```
+
+_Example output of `noncompliant01.py`:_
+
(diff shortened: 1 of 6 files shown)


docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
```diff
@@ -0,0 +1,176 @@
...
+z = 2
+z *= 2 + 1
+print(f"z *= 2 + 1    ={z}")
+```
+
+If a method changes an object’s state (has side effects) and is called multiple times within one expression, the result can be surprising and incorrect. For further info on python's order of precedence refer to The Python Language Specification , §6.16, "Evaluation Order" [[PLR 2022](https://docs.python.org/3/reference/expressions.html#evaluation-order)].
+
+## Non-Compliant Code Example
+
+This `noncompliant01.py` code is expected to provide labels for numbers obfuscates the evaluation and logic.
```
BartKaras1128:
```suggestion
`noncompliant01.py` is expected to provide labels for numbers, but it unnecessarily obfuscates the evaluation and logic.
```

BartKaras1128: It didn't make much sense to me the way this was written previously, I think this way is nicer.

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/noncompliant01.py
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: MIT
+"""Non-Compliant Code Example"""
+
+
+def label(number: int) -> str:
+    a = int(number < 0)  # negative flag
+    b = (number & 1) ^ 1  # even flag (1 for even, 0 for odd)
+    c = int(number < 5)  # small flag
+
+    key = (a << 2) | (b << 1) | c  # pack flags into a single key
```
s19110: I think it's a better practice for the variable names to indicate what the flag means instead of using comments:
```suggestion
    is_negative = int(number < 0)
    is_even = (number & 1) ^ 1  # (1 for even, 0 for odd)
    is_small = int(number < 5)

    key = (is_negative << 2) | (is_even << 1) | is_small  # pack flags into a single key
```

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
```diff
@@ -0,0 +1,176 @@
...
+```python
+"""Non-Compliant Code Example"""
+
+
+def label(number: int) -> str:
+    a = int(number < 0)  # negative flag
+    b = (number & 1) ^ 1  # even flag (1 for even, 0 for odd)
+    c = int(number < 5)  # small flag
+
+    key = (a << 2) | (b << 1) | c  # pack flags into a single key
```
s19110: To match my suggestion from noncompliant01.py - variable names should tell us what the variable is used for:
```suggestion
    is_negative = int(number < 0)
    is_even = (number & 1) ^ 1  # (1 for even, 0 for odd)
    is_small = int(number < 5)

    key = (is_negative << 2) | (is_even << 1) | is_small  # pack flags into a single key
```

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/compliant01.py
```diff
@@ -0,0 +1,17 @@
...
+        return "neg"
+    if number % 2 == 0:
+        return "even"
+    if number < 5:
+        return "small"
+    return "big"
+
+
+for number in range(-3, 3):
+    print(f"{number} = {label(number)}")
```
s19110, line 20: Personally, it bugs me that only one label can be applied to a number at a time. For example, 2 is both even and small, yet it will only be labeled as "even". Choice between which label takes priority seems arbitrary. We could store the labels in a collection to avoid that. I chose a set since each label is unique and the order shouldn't matter:
```suggestion
"""Compliant Code Example"""


def label(number: int):
    labels = set()
    if number < 0:
        labels.add("neg")
    if number % 2 == 0:
        labels.add("even")
    if number < 5:
        labels.add("small")
    else:
        labels.add("big")    
    return labels


for number in range(-3, 3):
    print(f"{number} = {label(number)}")
```

Then again, noncompliant01.py doesn't take multiple labels into account either, and it's not directly related to the described vulnerability, so I'd like to leave this choice up to you.

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
```diff
@@ -0,0 +1,176 @@
...
+## Bibliography
+
+<table>
+    <tr>
+        <td>[Python docs 2025 - simple statements]</td>
+        <td>7.2.1. Augmented assignment statements [online]. Available from: <a href="https://docs.python.org/3/reference/simple_stmts.html?highlight=augmented%20assignment%20operators#augmented-assignment-statements">https://docs.python.org/3/reference/simple_stmts.html?highlight=augmented%20assignment%20operators#augmented-assignment-statement</a>,  [Accessed 19 September 2025]</td>
+    </tr>
+    <tr>
+        <td>[python power 2025]</td>
+        <td>7.2.1. Augmented assignment statements [online]. Available from: <a href="https://docs.python.org/3/reference/simple_stmts.html?highlight=augmented%20assignment%20operators#augmented-assignment-statements">https://docs.python.org/3/reference/simple_stmts.html?highlight=augmented%20assignment%20operators#augmented-assignment-statement</a>,  [Accessed 19 September 2025]</td>
```
s19110: Duplicated reference from above:
```suggestion
        <td>6. Expressions [online]. Available from: <a href="https://docs.python.org/3/reference/expressions.html#index-59">https://docs.python.org/3/reference/expressions.html#index-59</a>,  [Accessed 19 September 2025]</td>
```

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/README.md
```diff
@@ -0,0 +1,176 @@
...
+    </tr>
+    <tr>
+        <td>[PLR 2022]</td>
+        <td>6.16. Evaluation order [online]. Available from: <a href="https://docs.python.org/3/reference/expressions.html#evaluation-orde">https://docs.python.org/3/reference/expressions.html#evaluation-orde</a>,  [Accessed 19 September 2025]</td>
```
s19110: Broken link. We wouldn't want any LLMs to complain about not being able to add the 'r' at the end now, would we? :wink: 
```suggestion
        <td>6.16. Evaluation order [online]. Available from: <a href="https://docs.python.org/3/reference/expressions.html#evaluation-order">https://docs.python.org/3/reference/expressions.html#evaluation-order</a>,  [Accessed 19 September 2025]</td>
```

docs/Secure-Coding-Guide-for-Python/CWE-691/CWE-783/compliant01.py
myteron: Have not tried but we could potentially update the `noncompliant01.py` to support multiple labels.

Issue #714, opened 2024-12-17
pySCG: Either add is() to CWE-595 or create a new rule


myteron:
While creating https://github.com/ossf/wg-best-practices-os-developers/tree/main/docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595 we found there is a need to explain why we need to distinguish between `==` and `is()`.  CWE-595 does not explain is().

Either add is() to CWE-595  or create a new rule.

s19110:
I can look into this.

PR #997, opened 2025-09-26
pySCG: Adding explanation of the 'is' operator to CWE-595

s19110:
Addresses issue [#714](https://github.com/ossf/wg-best-practices-os-developers/issues/714)

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
@@ -2,7 +2,7 @@
 
 Prevent unexpected results by knowing the differences between comparison operators such as `==` and `is`.
 
-In Python, the `==` operator is implemented by the `__eq__` method on an object [[python.org data model 2023](https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__eq__)]. For built-in types like `int` and `str`, the comparison is implemented in the interpreter. The main issue comes when implementing custom classes, where the default implementation compares object references using the `is` operator. The `is` operator compares the identities of the objects, equivalent to `id(obj1) == id(obj2)`. 
+In Python, the `==` operator is implemented by the `__eq__` method on an object [[python.org data model 2023](https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__eq__)]. For built-in types like `int` and `str`, the comparison is implemented in the interpreter. The main issue comes when implementing custom classes, where the default implementation compares object references using the `is` operator. The `is` operator compares the identities of the objects, equivalent to `id(obj1) == id(obj2)`.
 In CPython, this is their memory address. Everything in Python is an object, and each object is stored at a specific memory location [[de Langen 2023](https://realpython.com/python-is-identity-vs-equality/)].
 
 You want to implement the `__eq__` method on a class if you believe you ever want to compare it to another object or find it in a list of objects. Actually, it is so common that the `dataclasses.dataclass` decorator by default implements it for you [[dataclasses — Data Classes — Python 3.11.4 documentation](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass)].


docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
```diff
@@ -6,7 +6,7 @@ You want to implement the `__eq__` method on a class if you believe you ever wan
 
 ## Non-Compliant Code Example
 
-The non-compliant code shows how the default comparison operator compares object references rather than the object values. Furthermore, it displays how this causes issues when comparing lists of objects, although it applies to other types of collections as well. Finally, it shows how the `in` operator also depends on the behavior of the `__eq__` method and, therefore, also returns a non-desirable result.
+The non-compliant code shows how the default comparison operator compares object references rather than the object values. Furthermore, it displays how this causes issues when comparing lists of objects, although it applies to other types of collections as well. Then, it shows how the `in` operator also depends on the behavior of the `__eq__` method and, therefore, also returns a non-desirable result. Finally, it performs the comparison with the `is` operator, which checks as to whether the references point to the same object regardless of the stored value.
```
myteron: Some suggestion, however I still want to figure out the side effects when dealing with memory optimization in Python such as numbers between `-5 to 256`. There are similar issues when dealing with strings.

```suggestion
The `noncompliant01.py` code demonstrates potentially unexpected outcomes when using different comparisons. 

* The `==` operator using `__eq__`, checks value equality for most build-in types, checks for reference equality if the `__eq__` is missing  in a custom class. So 12 == 12 is True and Integer(12) == Integer(12) is False.
* The `==` comparing lists of objects, that also applies to other types of collections. 
* The `in` operator also depends on the behavior of the `__eq__` method
* The `is` operator that checks the references point to the same object regardless of the stored value.

```

docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
myteron, line 1: Suggstion for first sentence:

Prevent unexpected results by knowing what comparisment operators do such as `==` and `is` actually do.

After the existing first sentence we shall also explain:

Python falls back to comparing objects `id()` if the `__eq__ implementation is missing for a custom class.

docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
```diff
@@ -6,7 +6,7 @@ You want to implement the `__eq__` method on a class if you believe you ever wan
```
myteron, line 9: Be aware of Python's memory optimization for strings and numbers as demonstrated in `example01.py` code.
Python tries to avoid allocating more memory for the same string. Numbers `-5` to `256` are so frequently used that they are pre-reserved.

```py
""" Code Example """

print("-" * 10 + "Memory optimization with strings" + 10 * "-")
a = "foobar"
b = "foobar"
c = ''.join(["foo", "bar"])
print(f"a is b: {a} is {b}?", a is b)
print(f"a is c: {a} is {c}?", a is c)
print(f"a == c: {a} == {c}?", a == c)
print(f"size? len(a)={len(a)} len(b)={len(b)} len(c)={len(c)}")

print("-" * 10 + "Memory optimization with numbers" + 10 * "-")
a = b = 256 
print (f"{a} is {b}?", a is b)
a = b = 257
print (f"{a} is {b}?", a is b)

print("-" * 10 + "Memory optimization with numbers in a loop" + 10 * "-")
a = b = 255
while(a is b):
    a += 1 
    b += 1
    print (f"{a} is {b}?", a is b)
```

__The `example01.py` code output__:

```bash
a is b: foobar is foobar? True
a is c: foobar is foobar? False
a == c: foobar == foobar? True
size? len(a)=6 len(b)=6 len(c)=6
----------Memory optimisation with numbers----------
256 is 256? True
257 is 257? True
----------Memory optimisation with numbers in a loop----------
256 is 256? True
257 is 257? False
```

The first print statement illustrates Python's memory optimization for strings. 
See `id(a)` and `id(b)` printing the same object number compared to `id{c}` printing a different id for t he same string.

The example in the middle creates same object number as `a = b = 257` tells python that it can used the same memory.

The last print statements in `example01.py` illustrate Python's memory optimization for numbers between `-5 to 256`. Python needs to allocate new objects for numbers greater then `256`. Note that the behavior of Python for numbers between `-5` and `256` change depending on how its run. Following code changes behavior when run via interactive Python shell or in a file:

__example02.py:__

```py
a = 256
b = 256
print(a is b)
a = 257
b = 257
print(a is b)
```
Using an interactive Python shell will print `True` and `False` while running the code in a python script will print `True` and `True` as the runtime has optimized the code to preserve memory.

myteron (commented):
provided example01.py code in the md for the obsevered issues around numbers and strings to explain memory optimistation related isses.

docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
```diff
@@ -1,14 +1,70 @@
...
+Prevent unexpected results by knowing the differences between comparison operators such as `==` and `is`.
+
+ Python falls back to comparing objects' `id()` if the `__eq__` implementation is missing for a custom class. In Python, the `==` operator is implemented by the `__eq__` method on an object [[python.org data model 2023](https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__eq__)]. For built-in types like `int` and `str`, the comparison is implemented in the interpreter. The main issue comes when implementing custom classes, where the default implementation compares object references using the `is` operator. The `is` operator compares the identities of the objects, equivalent to `id(obj1) == id(obj2)`. The `id` function is built into Python, and in the CPython interpreter, the standard implementation, it returns the object's memory address [[de Langen 2023](https://realpython.com/python-is-identity-vs-equality/)].
```
myteron: The last sentence is bit odd. 

Also de Langen does not say it like that, so I suggest to quote him directly. Alternative is to use the python.org documentation.

```suggestion
In Python, the `==` operator is implemented by the `__eq__` method on an object [[python.org data model 2023](https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__eq__)]. For built-in types like `int` and `str`, the comparison is implemented in the interpreter. The main issue comes when implementing custom classes, where the default implementation compares object references using the `is` operator. The `is` operator compares the identities of the objects, equivalent to `id(obj1) == id(obj2)`. 
In CPython, this is their memory address. Everything in Python is an object, and each object is stored at a specific memory location [[de Langen 2023](https://realpython.com/python-is-identity-vs-equality/)].
```

docs/Secure-Coding-Guide-for-Python/CWE-697/CWE-595/README.md
```diff
@@ -1,14 +1,70 @@
...
+
 ## Non-Compliant Code Example
 
-The non-compliant code shows how the default comparison operator compares object references rather than the object values. Furthermore, it displays how this causes issues when comparing lists of objects, although it applies to other types of collections as well. Finally, it shows how the `in` operator also depends on the behavior of the `__eq__` method and, therefore, also returns a non-desirable result.
+The `noncompliant01.py` code demonstrates potentially unexpected outcomes when using different comparisons.
 
-[*noncompliant01.py:*](noncompliant01.py)
+* The `==` operator using `__eq__`, checks value equality for most build-in types, checks for reference equality if the `__eq__` is missing  in a custom class. So 12 == 12 is True and Integer(12) == Integer(12) is False.
```
myteron:
```suggestion
* The `==` operator using `__eq__`, checks value equality for most build-in types, checks for reference equality if the `__eq__` is missing  in a custom class. So `12 == 12` is `True` and `Integer(12) == Integer(12)` is `False`.
```

myteron (changes requested):

PR #998, opened 2025-09-29
Optimize hardened compiler options parser.

sueszli:
Improvements:

- Using uv inline dependencies to make script standalone and runnable from anywhere without any additional packages.
- Code quality: More functional style. Fewer states and branches. Increased code locality.
- Trying to parse data from a file, but falling back to remote Github blob if not found.


thomasnyman:
@mayank-ramnani: Do you have interest in reviewing this?

gkunz:
One issue I noticed based on the fact that dependabot recently bumped one of the dependencies of the original scraper: dependendabot currently only support uv.lock files - and that apparently incorrectly - but not yet the dependencies embedded in the script header:

https://github.com/dependabot/dependabot-core/issues/11946
https://github.com/dependabot/dependabot-core/issues/10478

So, I am kind of torn about the move to uv now, unfortunately. Other opinions?

gkunz:
Hi @sueszli,

while it can be a bit of an opinionated debate whether or not to pin dependencies, I'd say that pinning and consciously updating them is a best practice we want to promote. Scorecard for instance checks for pinned dependencies, too.

I was playing with the PR and uv a bit. I very much like that uv takes care of the dependencies and environment management, so that's a great improvement. In order to keep pinned dependencies and make dependabot pick them up, we could use a pyproject.toml file (instead of the former requirements.txt) and maintain dependencies in there. uv reads this file automatically.

I realize that my suggestions backtracks a few of your changes (separate directory, lightweight dependency handling), but I hope you agree on the dependency management aspects.

I like the other changes.

Finally, once dependabot can handle dependencies defined in uv script metdata [1], I see no reason to move back to this even more lightweight implementation.

[1] https://github.com/dependabot/dependabot-core/issues/11946

docs/Compiler-Hardening-Guides/compiler-options-scraper/.gitignore
```diff
@@ -1 +1,216 @@
-compiler-options.json
+# Byte-compiled / optimized / DLL files
```
gkunz, line 1: This is a lot of content. Did you accidentally include your private .gitignore?

docs/Compiler-Hardening-Guides/compiler-options-scraper/README.md
```diff
@@ -1,17 +1,26 @@
...
+make run   # Execute the scraper (bootstrapping venv if required)
+```
+
+## Dependency Management
 
-## Dependencies
+Dependencies are effectively locked via `requirements.txt` to support Dependabot analysis (which currently lacks `uv.lock` parsing support).
 
-Dependencies are specified in `requirements.txt`. The main dependencies are:
+Do not manually edit `requirements.txt`. To modify dependencies:
```
gkunz: I am wondering about the statement to not modify `requirements.txt`: if we (or dependabot) intend to bump a dependency, `requirement.txt` would be the place to do that, right?

docs/Compiler-Hardening-Guides/compiler-options-scraper/.gitignore
sueszli: This is the default `.gitignore` template for python provided by GitHub.

gkunz: ah, ok. I wasn't aware. Keep it then.

docs/Compiler-Hardening-Guides/compiler-options-scraper/README.md
sueszli: I agree! Very good point.

sueszli:
Thank you for your feedback @gkunz 

> This also broke my gh cli.

Whoops - Sorry about that.
I had to re-sign all commits, rebase and force-push which made a mess.

I'll squash redudnant commits now.
Looking forward to the next review cycle.
Thank you for your time!

docs/Compiler-Hardening-Guides/compiler-options-scraper/README.md
```diff
@@ -1,17 +1,25 @@
...
+The scraper normalizes content from the [OSSF Compiler Options Hardening Guide for C/C++](https://raw.githubusercontent.com/ossf/wg-best-practices-os-developers/refs/heads/main/docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C%2B%2B.md).
 
 ## Usage
 
-`python3 main.py`
+The project utilizes a `Makefile` for standard operations.
+
+```bash
+make help  # Show all available commands
+make run   # Execute the scraper (bootstrapping venv if required)
```
mayank-ramnani, line 15: Nitpick: If we're using make run to run the scraper, we should somehow ensure the commands run by make itself are not part of the output json file. 
```
╰─ make run > output.json && head output.json
test -d .venv || (uv venv .venv --python 3.13 && uv pip install -r requirements.txt)
uv run main.py
{
    "version": "2025-11-11",
    "options": {
        "recommended": [
```

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
```diff
@@ -1,149 +1,91 @@
...
+    with urllib.request.urlopen(fallback) as response:
+        assert response.getcode() == 200
+        return response.read().decode("utf-8")
 
 
 if __name__ == "__main__":
-    main()
+    content = get_content()
+
+    html = markdown.markdown(content, extensions=["tables"])
```
mayank-ramnani, line 79: Why are we adding a new dependency on the markdown package?
Instead of directly using the [HTML generated from the website](
https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html), we are now downloading the markdown source (from either local file system or github servers), rendering it into HTML using `markdown.markdown` and then parsing through it. 
Seems like an additional step and the benefit is unclear.

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
```diff
@@ -1,149 +1,91 @@
...
+    description, prerequisite = get_desc_preq_pair(row_data["Description"])
+
+    option_dict = {
+        "opt": row_data["Compiler Flag"],
+        "description": description,
+        "requires": extract_versions(row_data["Supported since"]),
+    }
 
+    if prerequisite:
+        option_dict["prerequisite"] = prerequisite
```
mayank-ramnani, line 42: (Nit) New change: `prereq` -> `prerequisite`

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
```diff
@@ -1,149 +1,91 @@
...
+    split_index = desc.find("Requires")
+    return (desc[:split_index], desc[split_index:]) if split_index != -1 else (desc, "")
+
+
+def create_option_dict(row_data: Dict[str, str]) -> Dict[str, Any]:
+    description, prerequisite = get_desc_preq_pair(row_data["Description"])
+
+    option_dict = {
+        "opt": row_data["Compiler Flag"],
+        "description": description,
```
mayank-ramnani, line 37: (Nit) Note that previously this field was called "desc" and is now being changed to "description".
I don't mind the increase in verbosity change, but please be mindful that it may break dependent tools since we are changing the structure.
Will let others opine on this.

mayank-ramnani (commented):
Hi Yahya,
I've reviewed the latest changes.
I have added 1 implementation related question and a few nitpick comments.
Thanks

docs/Compiler-Hardening-Guides/compiler-options-scraper/README.md
sueszli: Done: https://github.com/ossf/wg-best-practices-os-developers/pull/998/commits/41b20eaf32f31271a5308d32d2bbc26214e2fc85

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
sueszli: The Markdown file is the actual source of truth. The HTML we see on the website is generated by GitHub using an internal static-site pipeline, so consuming it directly means relying on opaque, derived output rather than the authored content.

We convert Markdown to HTML because the HTML tooling ecosystem is more mature. In particular, libraries like BeautifulSoup are far more robust for traversal and extraction than any Markdown AST tooling I am aware of, so this conversion step simplifies the implementation while improving reliability. Happy to modify if you have any recommendations for MD parsers.

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
sueszli: Sorry, I didn't catch that. Did you mean to change `prereq` to `prerequisite` or vice versa? The code currently uses `prerequisite`. Is this a necessary change?

docs/Compiler-Hardening-Guides/compiler-options-scraper/main.py
sueszli: Makes sense.
I also believe the JSON structure could be simplified further.
Not sure what the consensus is on this.

PR #999, opened 2025-10-01
Add CMake reference implementation of hardening guide

sueszli:
> Builds on top of: https://github.com/ossf/wg-best-practices-os-developers/pull/998

This commit translates the recommendations `from Compiler-Options-Hardening-Guide-for-C-and-C++.md`
into a universal, script-like CMake configuration that serves as a plug-and-play compiler 
hardening template.

Purpose:

Rather than manually selecting and applying flags from the guide, this template automatically
applies the appropriate recommendations based on:

- Compiler type (GCC vs Clang)
- Operating system (Linux-specific protections)
- Architecture (x86_64 vs ARM64)
- Build type (Release vs Debug)

This improves accessibility by letting developers drop the configuration into their projects and
immediately benefit from the guide's recommendations without needing expertise in compiler
security flags.

Implementation:

- Enforces latest stable compiler versions (GCC ≥14.0.0, Clang ≥16.0.0)
- Uses conditional logic to apply the correct flags for each environment
- Includes Dockerfile and Makefile for testing with both compilers

Directory Restructure:

- `example-minimal/`: Original Makefile approach
- `example-cmake/`: New universal template approach

This transforms the static guide into an actionable, copy-paste template that makes compiler
hardening accessible regardless of security expertise.

### DIFF
--- a/docs/Compiler-Hardening-Guides/example-cmake/Makefile
+++ b/docs/Compiler-Hardening-Guides/example-cmake/Makefile
@@ -1,4 +1,4 @@
-DOCKER_RUN = docker run --rm -v $(PWD):/workspace compiler-hardening-example-cmake sh -c
+DOCKER_RUN = docker run --rm -v $(PWD):/workspace:ro compiler-hardening-example-cmake sh -c
 
 .PHONY: require-docker # Ensure Docker is installed and running
 require-docker:


sueszli:
Needs `Product: Compiler Hardening Guide` label :)

sueszli:
@gkunz just checking in :)

docs/Compiler-Hardening-Guides/example-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+endif()
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|arm64|ARM64")
+  add_compile_options(-mbranch-protection=standard)
+endif()
+
+#
+# sources
+#
+
+add_executable(demo demo.c)
```
06kellyjac: missing newline EOF

Image

docs/Compiler-Hardening-Guides/example-cmake/Makefile
```diff
@@ -0,0 +1,17 @@
...
+run-gcc: build-image
+	$(DOCKER_RUN) "mkdir -p /tmp/gcc-build && cd /tmp/gcc-build && cmake -DCMAKE_C_COMPILER=gcc /workspace && make && ./demo"
+
+.PHONY: run-clang
+run-clang: build-image
+	$(DOCKER_RUN) "mkdir -p /tmp/clang-build && cd /tmp/clang-build && cmake -DCMAKE_C_COMPILER=clang /workspace && make && ./demo"
+
+.PHONY: clean
+clean:
+	docker rmi main
```
06kellyjac:
```suggestion
DOCKER_RUN = docker run --rm -v $(PWD):/workspace compiler-hardening-example-cmake sh -c

.PHONY: build-image
build-image:
	docker build -t compiler-hardening-example-cmake .

.PHONY: run-gcc
run-gcc: build-image
	$(DOCKER_RUN) "mkdir -p /tmp/gcc-build && cd /tmp/gcc-build && cmake -DCMAKE_C_COMPILER=gcc /workspace && make && ./demo"

.PHONY: run-clang
run-clang: build-image
	$(DOCKER_RUN) "mkdir -p /tmp/clang-build && cd /tmp/clang-build && cmake -DCMAKE_C_COMPILER=clang /workspace && make && ./demo"

.PHONY: clean
clean:
	docker rmi compiler-hardening-example-cmake
```

More understandable image name

docs/Compiler-Hardening-Guides/example-cmake/Makefile
```diff
@@ -0,0 +1,17 @@
...
+build-image:
+	docker build -t main .
+
+.PHONY: run-gcc
+run-gcc: build-image
+	$(DOCKER_RUN) "mkdir -p /tmp/gcc-build && cd /tmp/gcc-build && cmake -DCMAKE_C_COMPILER=gcc /workspace && make && ./demo"
+
+.PHONY: run-clang
+run-clang: build-image
+	$(DOCKER_RUN) "mkdir -p /tmp/clang-build && cd /tmp/clang-build && cmake -DCMAKE_C_COMPILER=clang /workspace && make && ./demo"
```
06kellyjac: Should probably have a `README.md` explaining the requirement on `docker` & that these commands will compile & run the demo within a container

docs/Compiler-Hardening-Guides/example-cmake/Makefile
```diff
@@ -0,0 +1,17 @@
+DOCKER_RUN = docker run --rm -v $(PWD):/workspace main sh -c
```
06kellyjac:
```suggestion
DOCKER_RUN = docker run --rm -v $(PWD):/workspace:ro main sh -c
```

Since the build is done inside `/tmp` we can safely set workspace to readonly 🙂

docs/Compiler-Hardening-Guides/example-cmake/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM --platform=linux/amd64 ubuntu:24.04
```
06kellyjac, line 1:
```suggestion
FROM ubuntu:24.04
```

Is it necessary to force `linux/amd64`?

docs/Compiler-Hardening-Guides/get_options.py
```diff
@@ -0,0 +1,95 @@
+#
```
06kellyjac, line 1: Probably want to land the python changes PR before merging this PR

docs/Compiler-Hardening-Guides/example-cmake/Makefile
sueszli: Oh I didn't know about this. Clever, thank you!

docs/Compiler-Hardening-Guides/example-cmake/Dockerfile
sueszli: This is actually required to enable x86 instruction set emulation via QEMU (through binfmt_misc), which Docker relies on to run x86 images on an ARM system.

docs/Compiler-Hardening-Guides/get_options.py
sueszli: I agree. This PR shouldn't be merged before https://github.com/ossf/wg-best-practices-os-developers/pull/998

Issue #1028, opened 2026-01-26
pyscg: missing to cover non-numeric/different types of rounding in 0005


myteron:
While working on 0005 "Specify Rounding for Numeric Conversions" during #1017 we found that there a different types of rounding. Some do not even have to do with numbers. This investigative issue is to seek out a list of rounding, specifically those non-numeric or on special numbers.

This issue is to find out other types of rounding and to decide to either just mention them in 0005 or to create an individual rule.

s19110:
I'm going to investigate this.

s19110:
For numbers, rounding in Python is delegated to the `__round__` special method: https://docs.python.org/3/library/functions.html#round

From what I managed to find, it is implemented in CPython for:

- int: https://github.com/python/cpython/blob/6611f4df80bbceb4b2fdf9d132c7652eacbd7743/Objects/longobject.c#L6202
- bool - it inherits from `PyLong_Type` and thus inherits the `__round__` method from `int`. This results in `round(True)` returning `1` and `round(False)` returning `0`
- float: https://github.com/python/cpython/blob/6611f4df80bbceb4b2fdf9d132c7652eacbd7743/Objects/floatobject.c#L1024
- decimal.Decimal: https://github.com/python/cpython/blob/6611f4df80bbceb4b2fdf9d132c7652eacbd7743/Modules/_decimal/_decimal.c#L4301
- fractions.Fraction: this one implements `__round__` in Python: https://github.com/python/cpython/blob/6611f4df80bbceb4b2fdf9d132c7652eacbd7743/Lib/fractions.py#L970
- unittest.mock.MagicMock: https://docs.python.org/3/library/unittest.mock.html#magicmock-and-magic-method-support - this class doesn't implement `__round__` per se but it allows you to mock its return value, for example like so:
```
from unittest.mock import MagicMock

m = MagicMock()
m.__round__.return_value = 42

print(round(m))       # 42
print(m.__round__.call_count)  # 1
```

Apart from that, even though `complex` could be considered a numeric type, it doesn't implement `__round__` and will throw `TypeError: type complex doesn't define __round__ method`. Instead, you should round specifically the real or imaginary part of the complex number.


I think explaining non-numeric rounding in pyscg-0005 might go beyond the scope of this rule because any class can support rounding as long as it implements `__round__`. I think it would be valuable to mention the `__round__` magic method in the rule but focus on the numeric examples that already exist in Python.

Please, let me know what's your thoughts on this.

[...]

Issue #1027, opened 2026-01-26
pyscg: Reduce 0003 content to be more concise


myteron:
0053 has been merged into 0003 as part of #1017.
Optimizing 0003 is to much for the PR that is updating 03_numbers. This issue is to reduce 0003 to something more useful.

as per @s19110 comment in PR: "Maybe we could add noncompliant01.py and create a corresponding compliant01.py"

BartKaras1128:
I'll start investigating this!

Issue #1017, opened 2025-12-22
pyscg: assure titles match content, reduce title length, reduce duplication, review content and flow


myteron:
Title renaming is to address a comment by the MITRE CWE team. 
New titles and TODO's have been drafted on a branch in a fork:
https://github.com/myteron/wg-best-practices-os-developers/blob/pyscg_titles_1017/docs/Secure-Coding-Guide-for-Python/readme_draft.md

While looking through the title's as part of #1009 we found some issues:
- First paragraphs lead to misinterpretation and need to be corrected, such as pyscg-0040
- Contradicting content such as 0053 vs 0003 "Avoid using bit shifts for Arithmetic"
- Misplaced content such as 0043 discussing Locale as main subject but also containing a "encoding" example that needs to move to 0045
- Potentially duplicated content.

PR's are organized by chapter.
Similar content is now much closer to each other we now see duplication that resulted in merging some of the rules. Addressing all of the issue's is turned out to cause delay's in PR per section so we created issue's to optimize these merges or fix bugs later. 

__Progress:__

01_introduction ~#1061~ / merged
02_encoding_and_strings  ~#1026~ / merged
03_numbers ~#1032~ / merged
04_neutralization ~#1044~ / merged
05_exception_handling ~#1043~ / merged
06_logging ~#1063 ~ / merged
07_concurrency ~#1073 ~ / merged 
08_coding_standards ~#1060~ / merged
09_cryptography ~#1057~ / merged

s19110:
PR for 07_concurrency is ready for review: https://github.com/ossf/wg-best-practices-os-developers/pull/1073/changes

PR #1029, opened 2026-01-26
pyscg: 03_numbers: applied title changes for 03_numbers. Merged 0053 into 0003

myteron:
as part of #1017  we had decided on new titles.

0001 was renamed to "Control Numeric Precision". Wording of Real vs other types of numbers corrected.
0002 renamed to "Guard C-Backed Numbers Against Overflow". C-Backed numbers explained. 0003 was updated to use new wording.
0003 renamed to "Use Arithmetic Over Bitwise Operations". Added 0053 content. Removed loop example (already covered in 0004). Created issue #1027 to reduce size of merged 0003 as a follow up.
0053 was merged into 0003 and 0053 removed.
0004 renamed to "Use Integer Loop Counters"
0005 renamed to "Specify Rounding for Numeric Conversions". added issue #1028 
0006 renamed to "Use an Appropriate Comparitor for Numbers" updated first sentence.
0007 renamed to "Use String Literals for Decimal Construction"

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0041/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0041/README.md
@@ -138,7 +138,7 @@ class TestSimulateDeployingFrontEnd(unittest.TestCase):
         config["LOGGING"] = {
             "level": "DEBUG",
         }
-        self.config_file_path = Path("config.ini", exist_ok=True)
+        self.config_file_path = Path("config.ini")
         with open(self.config_file_path, "w", encoding="utf-8") as config_file:
             config.write(config_file)
         self.config_file_path.chmod(0o400)
@@ -149,6 +149,7 @@ class TestSimulateDeployingFrontEnd(unittest.TestCase):
 
     def tearDown(self):
         """Clean up after us and remove the config file"""
+        self.config_file_path.chmod(0o600)
         self.config_file_path.unlink()
 
 
--- a/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0041/compliant01.py
+++ b/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0041/compliant01.py
@@ -42,7 +42,7 @@ class TestSimulateDeployingFrontEnd(unittest.TestCase):
         config["LOGGING"] = {
             "level": "DEBUG",
         }
-        self.config_file_path = Path("config.ini", exist_ok=True)
+        self.config_file_path = Path("config.ini")
         with open(self.config_file_path, "w", encoding="utf-8") as config_file:
             config.write(config_file)
         self.config_file_path.chmod(0o400)
@@ -53,6 +53,7 @@ class TestSimulateDeployingFrontEnd(unittest.TestCase):
 
     def tearDown(self):
         """Clean up after us and remove the config file"""
+        self.config_file_path.chmod(0o600)
         self.config_file_path.unlink()
 
 


docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0006/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0006: Incorrect Conversion Between Numeric Types
+# pyscg-0006: Use an Appropriate Comparator for Numbers
 
-String representations of floating-point numbers must not be compared or inspected outside of specialized modules such as `decimal` or `math`.
+Use the appropriate numeric type/module such as decimal, fractions, `math.isclose()` instead of string comparators.
```
tommcd: `decimal`, `fractions`, and `math.isclose()`

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0001/README.md
```diff
@@ -1,8 +1,8 @@
...
+# pyscg-0001: Control Numeric Precision
 
 Avoid floating-point and use integers or the `decimal` module to ensure precision in applications that require high accuracy, such as in financial or banking computations.
 
-In Python, floating-point types are constrained by a fixed number of binary mantissa bits, typically allowing for around seven decimal digits of precision (24-bit values). Consequently, they are not well-suited for representing surds, such as `√7` or `π` with full accuracy. Additionally, due to their binary nature, floating-point types are incapable of exactly terminating decimals in `base 10`, such as `0.3`, which has a repeating binary representation.
+In Python, floating-point types have limited precision, typically accurate to about seven decimal digits. Consequently, they cannot represent many numbers with full accuracy. For example, irrational numbers such as `√7` or `π` cannot be represented exactly. Additionally, due to their binary nature, floating-point types are incapable of exactly representing some terminating decimals in `base 10`, such as `0.3`, which has a repeating binary representation.
```
tommcd: Suggestions:
“about seven decimal digits” is true for 32-bit float (float32), but Python’s built-in float is almost always an IEEE-754 64-bit binary float, accurate to about 15–17 significant decimal digits.

The “limited precision” is significant digits, not “decimal digits after the point”.

`base 10` I wouldn't have backticks here as it's just normal prose.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0002/README.md
```diff
@@ -1,14 +1,14 @@
...
 - Built-in types such as `int`, `float`, or `complex` [[Python 2024]](https://docs.python.org/3.9/library/stdtypes.html). These types are provided by classes and are protected against overflows.
 
-- Primitive types share issues known from `C` , or `C++` and appear in `Python` to:
+- C-backed numbers are numeric types implemented in C (rather than pure Python) for performance or system compatibility. These primitive types share issues known from `C` or `C++` and appear in `Python` to:
   - interact with the operating system modules such as `time`.
-  - to be memory efficiency using modules such as `numpy` or `ctype`.
+  - to be memory efficient using modules such as `numpy` or `ctypes`.
 
-Developers should follow the `C` guidelines when using or interacting with `C` type variables.
+Developers should follow the `C` guidelines when using or interacting with C-backed numbers.
```
tommcd: Suggested text:

```
pyscg-0002: Guard Fixed-Width Integers Against Overflow

Ensure that integer overflow is properly handled to avoid unexpected behavior.

Python numeric values fall into two groups:

- **Python numeric objects** (`int`, `float`, `complex`) [[Python stdtypes]](https://docs.python.org/3/library/stdtypes.html):
  - `int` has arbitrary precision (bounded only by memory), so it cannot overflow by wrapping.
  - `float` and `complex` are IEEE-754 based with finite range; they can overflow to infinity or lose precision, but do not wrap.

- **Fixed-width, C-compatible integers** (e.g., NumPy dtypes like `np.int32`, `ctypes.c_int32`, `struct`-packed integers). These exist to:
  - interface with native libraries and system APIs.
  - provide memory-efficient numeric arrays.

Fixed-width integers wrap silently on overflow, just as in C. When using them, choose explicit width/signedness, validate bounds before arithmetic or casting, and account for wraparound semantics.
```
Reason:
The original lumps float/complex with int as "protected against overflows," but only int is arbitrary-precision. And "C-backed" is possibly misleading since CPython implements everything in C anyway. What matters is whether the type has fixed-width wraparound semantics. This revision targets the actual problem: NumPy/ctypes/struct integers that wrap silently like C.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -1,13 +1,17 @@
-# pyscg-0003: Promote Readability and Compatibility by Using Mathematical Written Code with Arithmetic Operations Instead of Bit-wise Operations
+# pyscg-0003: Use Arithmetic Over Bitwise Operations
```
tommcd, line 1: May this can be a separate improvement, but I suggest:

# pyscg-0003: Separate Arithmetic and Bitwise Intent

Treat integer data as either a **numeric value** or a **bit collection**—not both. Use arithmetic operators for numeric calculations and bitwise operators for bit manipulation; mixing them on the same data obscures intent and invites mistakes.

**Rule:** If a value represents a magnitude, do not use `<<`/`>>` to implement multiplication/division. If a value represents a bitfield, do not apply arithmetic beyond offsetting/packing logic.

This aligns with SEI CERT NUM01-J [[SEI CERT JAVA 2024]](https://wiki.sei.cmu.edu/confluence/display/java/NUM01-J.+Do+not+perform+bitwise+and+arithmetic+operations+on+the+same+data). Prefer clear arithmetic (`5*x + 1`) over shift-based "optimizations" (`(x << 2) + x + 1`).

## Why this matters in Python

- **Readability:** mixed operations make it harder to see what checks are needed (range validation vs masking logic).
- **Precedence traps:** `+`/`-` bind tighter than `<<`/`>>`, so `1 << n + 1` means `1 << (n + 1)`. Use parentheses when shifts appear in compound expressions.
- **Sign-extension surprise:** right shift on negative values is arithmetic (sign-extending), not logical. People expecting zero-fill get caught out.
- **Unbounded growth:** Python `int` doesn't overflow, so `1 << n` in a loop can explode memory/performance.
- **Boundary code:** when working with fixed-width integers (`ctypes`, NumPy dtypes, packed binary formats), C/ABI-like fixed-width semantics apply—handle width, signedness, and wraparound explicitly. Consult CERT C guidance [[SEI CERT C 2025]](https://www.securecoding.cert.org/confluence/display/seccode/CERT+C+Coding+Standard).

## When bitwise is appropriate

Bitwise operations are fine when data is actually a bit collection: flags/masks, serialization, binary protocols, cryptographic primitives. Keep such code localized, document the assumed bit width, and prefer helper functions for packing/unpacking.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -74,6 +78,139 @@ foo << 32: type <class 'int'> 0b101000110100100000000000000000000000000000000
 foo << 64: type <class 'int'> 0b10100011010010000000000000000000000000000000000000000000000000000000000000000
 ```
 
+## Bit-shifting Positive vs Negative Numbers
```
tommcd, line 81: Suggestion for future improvement:

to_bytes(x.__sizeof__(), ...) is misleading: __sizeof__() is object memory footprint, not representation width; it forces large sign-extension padding (0x00/0xFF) and does not reflect CPython’s internal int storage.

“-2 ends with hex 0xFE or binary 1110” is inconsistent: 0xFE is 11111110;

The “left shift reaches/remains at zero” explanation is confusing: reaching 0 is due to right shifts; left shift only preserves 0 if the value is already 0.

positive_int <<= 1
positive_int <<= 1
positive_int <<= 1000000000

only works because positive_int  is already set to 0

PR #1031, opened 2026-01-27
adding title description to contrib guide

myteron:
as part of #1017 we want to add the definition of a good title to the contribution guide.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/CONTRIBUTING.md
+++ b/docs/Secure-Coding-Guide-for-Python/CONTRIBUTING.md
@@ -163,13 +163,11 @@ Example structure with mocked up data below `docs/Secure-Coding-Guide-for-Python
 
 ### Rule Titles
 
-Titles should be:
+Titles should:
 
-* Positive
-* Action-oriented guideline statements
-* APA Title Case style [[APA 2022](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case)]
-
-Using a negative form only when it reads more naturally.
+* Be written in APA Title Case style [[APA 2022](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case)]
+* Be action-oriented
+* Preferably describe wanted design over anti-pattern
 
 ### Coding Examples
 


docs/Secure-Coding-Guide-for-Python/CONTRIBUTING.md
```diff
@@ -161,6 +161,16 @@ Example structure with mocked up data below `docs/Secure-Coding-Guide-for-Python
...
 
+### Rule Titles
+
+Titles should be:
+
+* Positive
+* Action-oriented guideline statements
+* APA Title Case style [[APA 2022](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case)]
+
+Using a negative form only when it reads more naturally.
```
tommcd: As written, “Using a negative form only …” is a dangling participle: it doesn’t clearly attach to the implied subject (“Titles”), so it reads disconnected.

Suggestion: "Use" with a slight rewording.

“Use a negative form only when a positive form would be awkward or less clear.”

Or, bring it into the list:

Titles should be:
- Positive
- Action-oriented guideline statements
- Written in APA Title Case
- In negative form only when a positive form would be awkward or less clear

myteron: Well "In negative form only when a positive form would be awkward or less clear" is kind a inversion or softening of "Positive". So instead of excemptions on inversions we can also "soften" the preferred "Positive"

so I think we should remove the 'inverted' case/ last bullet point and replace the first instead.

Options (to replace "Positive"):

* Quote the recommended design 
* Prefer "do's" over "dont's"
* Emphasize guidance over prohibition
* Affirmative when possible
* Based on preferred design pattern

one of those?

tommcd: > Prefer “dos” over “don’ts”.  
Very clear, but informal, might suggest rules start with "Do"

I suggest:
"Prefer affirmative, action-oriented guidance over prohibitions."

Or shorter:

"Prefer affirmative guidance when possible."

myteron (commented):
what needs to be done?

Titles should:

* Be written in APA Title Case
* Be action-oriented
* Preferably describe wanted design over anti-pattern

PR #1032, opened 2026-01-27
applied title changes for 03_numbers. Merged 0053 into 0003 while exc…

myteron:
as part of https://github.com/ossf/wg-best-practices-os-developers/issues/1017 we had decided on new titles.

0001 was renamed to "Control Numeric Precision". Wording of Real vs other types of numbers corrected.
0002 renamed to "Guard C-Backed Numbers Against Overflow". C-Backed numbers explained. 0003 was updated to use new wording.
0003 renamed to "Use Arithmetic Over Bitwise Operations". Added 0053 content. Removed loop example (already covered in 0004). Created issue https://github.com/ossf/wg-best-practices-os-developers/issues/1027 to reduce size of merged 0003 as a follow up.
0053 was merged into 0003 and 0053 removed.
0004 renamed to "Use Integer Loop Counters"
0005 renamed to "Specify Rounding for Numeric Conversions". added issue https://github.com/ossf/wg-best-practices-os-developers/issues/1028
0006 renamed to "Use an Appropriate Comparitor for Numbers" updated first sentence.
0007 renamed to "Use String Literals for Decimal Construction"

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0002/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0002/README.md
@@ -1,14 +1,14 @@
-# pyscg-0002: Guard C-Backed Numbers Against Overflow
+# pyscg-0002: Guard Fixed-Width Numbers Against Overflow
 
 Ensure that integer overflow is properly handled in order to avoid unexpected behavior. Python data types can be divided into two categories:
 
 - Built-in types such as `int`, `float`, or `complex` [[Python 2024]](https://docs.python.org/3.9/library/stdtypes.html). These types are provided by classes and are protected against overflows.
 
-- C-backed numbers are numeric types implemented in C (rather than pure Python) for performance or system compatibility. These primitive types share issues known from `C` or `C++` and appear in `Python` to:
+- Fixed-Width numbers are numeric types implemented in C (rather than pure Python) for performance or system compatibility. These primitive types share issues known from `C` or `C++` and appear in `Python` to:
   - interact with the operating system modules such as `time`.
   - to be memory efficient using modules such as `numpy` or `ctypes`.
 
-Developers should follow the `C` guidelines when using or interacting with C-backed numbers.
+Developers should follow the `C` guidelines when using or interacting with Fixed-Width numbers.
 
 ## Non-Compliant Code Example -- Using numpy.int64
 
@@ -213,7 +213,7 @@ The `compliant02.py` example is protecting the lower level c-lib from an `Overfl
 
 ## Non-Compliant Code Example -- Using math.exp
 
-The `noncompliant03.py` code example results in a `OverflowError: math range error`. This is due to `math.exp` being a `C` implementation behind the scenes for better performance. So while it returns a `Python float` it does use C-backed numbers internally for the calculation in `mathmodule.c` [[cpython 2024]](https://github.com/python/cpython/blob/main/Modules/mathmodule.c).
+The `noncompliant03.py` code example results in a `OverflowError: math range error`. This is due to `math.exp` being a `C` implementation behind the scenes for better performance. So while it returns a `Python float` it does Fixed-Width nubers internally for the calculation in `mathmodule.c` [[cpython 2024]](https://github.com/python/cpython/blob/main/Modules/mathmodule.c).
 
 *[noncompliant03.py](noncompliant03.py):*
 
--- a/docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
@@ -11,7 +11,7 @@ There are two `C` and `C++` design patterns for resource optimisation that are n
 
 The use of bit-wise operations for arithmetic or flag registers can reduce readability, predictability of the code and can also cause compatibility issues. Some bit-wise operations can reduce performance. Python tries to safeguard changes between positive and negative numbers by storing the sign separately. It tries to prevent overflows by using either `32-bit unsigned integer` arrays with `30-bit` digits or `16-bit` unsigned integer arrays with `15-bit` digit [[Rusher 2017]](https://rushter.com/blog/python-integer-implementation/). In other words, Python changes and adapts on the fly.
 
-A need to use bit-wise operations in Python can be due to translations or dealings with `C`, `C++` or `Java`, system libraries, raw binary data, or cryptographic algorithms. Existing Python modules hooking into system `C` libraries for cryptographic functions or math all to avoid the need to implement bit-shifting on a Python level. Bit-shifting can have unexpected outcomes. Python's ctypes module allows integration of `C` based system libraries into Python and direct access to C-backed numbers that can have different behavior than using high-level Python. Understanding `ctypes` or `C` requires understanding the *CERT C Coding Standard* [[SEI CERT C 2025]](https://www.securecoding.cert.org/confluence/display/seccode/CERT+C+Coding+Standard) and setting boundaries manually in Python.
+A need to use bit-wise operations in Python can be due to translations or dealings with `C`, `C++` or `Java`, system libraries, raw binary data, or cryptographic algorithms. Existing Python modules hooking into system `C` libraries for cryptographic functions or math all to avoid the need to implement bit-shifting on a Python level. Bit-shifting can have unexpected outcomes. Python's ctypes module allows integration of `C` based system libraries into Python and direct access to Fixed-Width numbers that can have different behavior than using high-level Python. Understanding `ctypes` or `C` requires understanding the *CERT C Coding Standard* [[SEI CERT C 2025]](https://www.securecoding.cert.org/confluence/display/seccode/CERT+C+Coding+Standard) and setting boundaries manually in Python.
 
 The `example01.py` code demonstrates bit-wise operators available in Python.
 
@@ -241,7 +241,7 @@ The statement in `compliant01.py` clarifies the programmer's intention.
 print(8 * 4 + 10)
 ```
 
-It is recommended by *[pyscg-0002: Guard C-Backed Numbers Against Overflow](../pyscg-0002/README.md)* to also check for under or overflow.
+It is recommended by *[pyscg-0002: Guard Fixed-Width Numbers Against Overflow](../pyscg-0002/README.md)* to also check for under or overflow.
 
 ## Non-compliant Code Example (Right Shift)
 
--- a/docs/Secure-Coding-Guide-for-Python/readme.md
+++ b/docs/Secure-Coding-Guide-for-Python/readme.md
@@ -119,7 +119,7 @@ It is __not production code__ and requires code-style or python best practices t
 </tr>
 
 <tr>
-<td><a href="03_numbers/pyscg-0002/README.md">pyscg-0002: Guard C-Backed Numbers Against Overflow</a></td>
+<td><a href="03_numbers/pyscg-0002/README.md">pyscg-0002: Guard Fixed-Width Numbers Against Overflow</a></td>
 <td></td>
 <td><a href="https://cwe.mitre.org/data/definitions/191.html">CWE-191</a>, <a href="https://cwe.mitre.org/data/definitions/190.html">CWE-190</a></td>
 </tr>


docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0001/README.md
```diff
@@ -1,8 +1,8 @@
...
+# pyscg-0001: Control Numeric Precision
 
 Avoid floating-point and use integers or the `decimal` module to ensure precision in applications that require high accuracy, such as in financial or banking computations.
 
-In Python, floating-point types are constrained by a fixed number of binary mantissa bits, typically allowing for around seven decimal digits of precision (24-bit values). Consequently, they are not well-suited for representing surds, such as `√7` or `π` with full accuracy. Additionally, due to their binary nature, floating-point types are incapable of exactly terminating decimals in `base 10`, such as `0.3`, which has a repeating binary representation.
+In Python, floating-point types have limited precision, typically accurate to about seven decimal digits. Consequently, they cannot represent many numbers with full accuracy. For example, irrational numbers such as `√7` or `π` cannot be represented exactly. Additionally, due to their binary nature, floating-point types are incapable of exactly representing some terminating decimals in `base 10`, such as `0.3`, which has a repeating binary representation.
```
s19110: While I was looking for the source of the info about the number of accurate decimal digits, I have found that since Python 3.11, CPython requires support for floating-point numbers in IEEE 745 standard: https://docs.python.org/3/whatsnew/3.11.html

I read that according to this standard, the single-precision float format (float32) offers 24 significant precision bits (which translates to about seven accurate digits) while double-precision (float64) offers 52 bits (about 15-16 accurate digits).

At first, I was confused where this number comes from. Perhaps we could keep the mention of the number of bits, specify which precision type is expected in our version of Python, and cite IEEE as our source?

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -1,13 +1,17 @@
-# pyscg-0003: Promote Readability and Compatibility by Using Mathematical Written Code with Arithmetic Operations Instead of Bit-wise Operations
+# pyscg-0003: Use Arithmetic Over Bitwise Operations
 
-Avoid using bitwise operations for calculations, write math as math instead to ensure code clarity, compatibility and maintainability.
+Avoid mixing bitwise shifts with arithmetic operations, instead, use clear mathematical expressions instead to maintain predictable behavior, readability, and compatibility
```
s19110: Missing dot at the end of the sentence:

```suggestion
Avoid mixing bitwise shifts with arithmetic operations, instead, use clear mathematical expressions instead to maintain predictable behavior, readability, and compatibility.
```

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -74,6 +78,139 @@ foo << 32: type <class 'int'> 0b101000110100100000000000000000000000000000000
 foo << 64: type <class 'int'> 0b10100011010010000000000000000000000000000000000000000000000000000000000000000
 ```
 
+## Bit-shifting Positive vs Negative Numbers
+
+Python tries to avoid issues around signed numbers by storing the sign in a separate data field. Bit-shifting a whole negative number will always remain negative in Python. Positive whole numbers will remain positive in Python or zero. Shifting a positive whole number to the right will never go below zero and shifting a negative number to the left will never go above minus one. Shifting to the left behaves differently between positive and negative numbers.
```
s19110, line 83: I see that the non-compliant code example from pyscg-0053 has been removed and wasn't transfered to this rule. Was it done to avoid this rule being too long? I thought it illustrated the difference between bit-shifting positive and negative numbers pretty nicely.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0006/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0006: Incorrect Conversion Between Numeric Types
+# pyscg-0006: Use an Appropriate Comparator for Numbers
```
s19110, line 1: Not directly related to the pull request, but in the title changes draft I saw a mention of creating an invstigative issue regarding non-numeric types our rounding. Has this issue been created? I couldn't find it on the project board.

s19110 (changes requested):
I left a couple of questions and suggestions.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0006/README.md
```diff
@@ -1,6 +1,6 @@
...
 
-String representations of floating-point numbers must not be compared or inspected outside of specialized modules such as `decimal` or `math`.
+Use the appropriate numeric type/module such as decimal, fractions, `math.isclose()` instead of string comparators.
```
myteron: @tommcd : you had commented "decimal, fractions, and math.isclose()" 

Should we provide a full list? "such as" is academic speak for "for example".

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0001/README.md
myteron: Comment from @tommcd was: 

----
Suggestions:
“about seven decimal digits” is true for 32-bit float (float32), but Python’s built-in float is almost always an IEEE-754 64-bit binary float, accurate to about 15–17 significant decimal digits.

The “limited precision” is significant digits, not “decimal digits after the point”.

base 10 I wouldn't have back ticks here as it's just normal prose.

----

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0002/README.md
```diff
@@ -1,14 +1,14 @@
...
 - Built-in types such as `int`, `float`, or `complex` [[Python 2024]](https://docs.python.org/3.9/library/stdtypes.html). These types are provided by classes and are protected against overflows.
 
-- Primitive types share issues known from `C` , or `C++` and appear in `Python` to:
+- C-backed numbers are numeric types implemented in C (rather than pure Python) for performance or system compatibility. These primitive types share issues known from `C` or `C++` and appear in `Python` to:
   - interact with the operating system modules such as `time`.
-  - to be memory efficiency using modules such as `numpy` or `ctype`.
+  - to be memory efficient using modules such as `numpy` or `ctypes`.
 
-Developers should follow the `C` guidelines when using or interacting with `C` type variables.
+Developers should follow the `C` guidelines when using or interacting with C-backed numbers.
```
myteron: comment from @tommcd :

Suggested text:

pyscg-0002: Guard Fixed-Width Integers Against Overflow

Ensure that integer overflow is properly handled to avoid unexpected behavior.

Python numeric values fall into two groups:

- **Python numeric objects** (`int`, `float`, `complex`) [[Python stdtypes]](https://docs.python.org/3/library/stdtypes.html):
  - `int` has arbitrary precision (bounded only by memory), so it cannot overflow by wrapping.
  - `float` and `complex` are IEEE-754 based with finite range; they can overflow to infinity or lose precision, but do not wrap.

- **Fixed-width, C-compatible integers** (e.g., NumPy dtypes like `np.int32`, `ctypes.c_int32`, `struct`-packed integers). These exist to:
  - interface with native libraries and system APIs.
  - provide memory-efficient numeric arrays.

Fixed-width integers wrap silently on overflow, just as in C. When using them, choose explicit width/signedness, validate bounds before arithmetic or casting, and account for wraparound semantics.

Reason:
The original lumps float/complex with int as "protected against overflows," but only int is arbitrary-precision. And "C-backed" is possibly misleading since CPython implements everything in C anyway. What matters is whether the type has fixed-width wraparound semantics. This revision targets the actual problem: NumPy/ctypes/struct integers that wrap silently like C.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -1,13 +1,17 @@
-# pyscg-0003: Promote Readability and Compatibility by Using Mathematical Written Code with Arithmetic Operations Instead of Bit-wise Operations
+# pyscg-0003: Use Arithmetic Over Bitwise Operations
```
myteron, line 1: from @tommcd :

May this can be a separate improvement, but I suggest:
pyscg-0003: Separate Arithmetic and Bitwise Intent

Treat integer data as either a numeric value or a bit collection—not both. Use arithmetic operators for numeric calculations and bitwise operators for bit manipulation; mixing them on the same data obscures intent and invites mistakes.

Rule: If a value represents a magnitude, do not use <</>> to implement multiplication/division. If a value represents a bitfield, do not apply arithmetic beyond offsetting/packing logic.

This aligns with SEI CERT NUM01-J [[SEI CERT JAVA 2024]](https://wiki.sei.cmu.edu/confluence/display/java/NUM01-J.+Do+not+perform+bitwise+and+arithmetic+operations+on+the+same+data). Prefer clear arithmetic (5*x + 1) over shift-based "optimizations" ((x << 2) + x + 1).
Why this matters in Python

    Readability: mixed operations make it harder to see what checks are needed (range validation vs masking logic).
    Precedence traps: +/- bind tighter than <</>>, so 1 << n + 1 means 1 << (n + 1). Use parentheses when shifts appear in compound expressions.
    Sign-extension surprise: right shift on negative values is arithmetic (sign-extending), not logical. People expecting zero-fill get caught out.
    Unbounded growth: Python int doesn't overflow, so 1 << n in a loop can explode memory/performance.
    Boundary code: when working with fixed-width integers (ctypes, NumPy dtypes, packed binary formats), C/ABI-like fixed-width semantics apply—handle width, signedness, and wraparound explicitly. Consult CERT C guidance [[SEI CERT C 2025]](https://www.securecoding.cert.org/confluence/display/seccode/CERT+C+Coding+Standard).

When bitwise is appropriate

Bitwise operations are fine when data is actually a bit collection: flags/masks, serialization, binary protocols, cryptographic primitives. Keep such code localized, document the assumed bit width, and prefer helper functions for packing/unpacking.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -74,6 +78,139 @@ foo << 32: type <class 'int'> 0b101000110100100000000000000000000000000000000
 foo << 64: type <class 'int'> 0b10100011010010000000000000000000000000000000000000000000000000000000000000000
 ```
 
+## Bit-shifting Positive vs Negative Numbers
```
myteron, line 81: comment from @tommcd :

In [docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md](https://github.com/ossf/wg-best-practices-os-developers/pull/1029#discussion_r2732088224):

> @@ -74,6 +78,139 @@ foo << 32: type <class 'int'> 0b101000110100100000000000000000000000000000000
 foo << 64: type <class 'int'> 0b10100011010010000000000000000000000000000000000000000000000000000000000000000
 ```
 
+## Bit-shifting Positive vs Negative Numbers

Suggestion for future improvement:

to_bytes(x.sizeof(), ...) is misleading: sizeof() is object memory footprint, not representation width; it forces large sign-extension padding (0x00/0xFF) and does not reflect CPython’s internal int storage.

“-2 ends with hex 0xFE or binary 1110” is inconsistent: 0xFE is 11111110;

The “left shift reaches/remains at zero” explanation is confusing: reaching 0 is due to right shifts; left shift only preserves 0 if the value is already 0.

positive_int <<= 1
positive_int <<= 1
positive_int <<= 1000000000

only works because positive_int is already set to 0

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0001/README.md
myteron: Hi @s19110 @thommcd, how about:


```suggestion
In Python, floating-point types have limited precision. A Python `float` has `~15–17` significant decimal digits total as per _Build Changes_ in _What's New in Python 3.11_ [python.org 2022](https://docs.python.org/3/whatsnew/3.11.html)— not `~15–17` digits after the decimal point.

Those digits are shared between:

* digits before the decimal point
* digits after the decimal point

So if the number gets bigger on the left, you automatically lose precision on the right.

Consequently, they cannot represent many numbers with full accuracy. For example, irrational numbers such as `√7` or `π` cannot be represented exactly. Additionally, due to their binary nature, floating-point types are incapable of exactly representing some terminating decimals in base 10, such as `0.3`, which has a repeating binary representation.
```

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0002/README.md
myteron: @tommcd: ok but we just changed "C-Type" numbers to "C-Backed" numbers. So I don't see how we can now change to "Fixed-Width" numbers? Also, this is not limited to Integers.

Or if "Fixed-Width" is the preferred wording then we should potentially also change the title. In practice we are only interested in describing and covering the use of "Python numeric objects" and tell the reader to read up secure coding of the language that provides Non-"Python numberic objects" when it comes to numbers. ctypes are however not limited to numbers.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
myteron: @tommcd

I have created issue  #1027  in order to cleanup 0003 after merging 0053 content into it. Please add comments for the cleanup in issue #1027 unless you find the current content unfit for release.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
myteron: I thought that this is already explained in at least one of the existing rules. There is also the issue that noncomplain01.py from 0053 is missing a compliant01.py counterpart. 
But we could also clean that up in the #1027 issue.

Let me know if I should add noncompliant01.py from 0053 into 0003 for now and leave cleaning up 0003 to the work in issue #1027

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0006/README.md
myteron: 0005 renamed to "Specify Rounding for Numeric Conversions". added issue #1028

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
myteron: @tommcd: please let me know if this work needs to be done prior to release or if I should create an issue to fix it later.

myteron: Discussed, Create an issue based on the commnent to fix 0003

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0006/README.md
myteron: discussed and resolved

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
myteron: added @tommcd to issue #1027

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
myteron: as discussed, added to issue on 0003 #1027  -resolved here.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0001/README.md
s19110: Looks good to me

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
s19110: I wasn't able to find it in any other rules besides 0053 and 0003. Maybe we could add noncompliant01.py and create a corresponding compliant01.py as part of the issue #1027 ?

PR #1035, opened 2026-01-29
Remove scorecard from WG projects

balteravishay:
Scorecard and Allstar have moved to the ORBIT WG. This PR removes the projects from the best practices projects list.


david-a-wheeler:
I fear that everyone assumed everyone had already talked.

My memory is imperfect, but I don't remember a formal "decision", I presumed that the people involved had agreed to it. If people haven't agreed to it, I suggest status quo until there's a discussion and agreement.

gkunz:
Thanks for the discussion. We have a Best WG call later today (3pm UTC) in which we can discuss this with the Scorecard project, if @spencerschrock and/or @justaugustus are available.

gkunz:
It was decided that Scorecard remains. Thus, I am closing this PR.

PR #1044, opened 2026-02-17
pyscg: 04_neutralization, renaming and organizing titles

BartKaras1128:
Renaming and organizing titles and rules in section 05_exception_handling part of #1071
as agreed in https://github.com/myteron/wg-best-practices-os-developers/blob/pyscg_titles_1017/docs/Secure-Coding-Guide-for-Python/readme_draft.md.

renamed title of 0008,0009,0010, 0012, 0013,0023,0047

Added a sentence to 0013 to explain "search path"
Going to create an issue on 0011 to Create issue (bug) that corrects the code examples for non-compliant and compliant code

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0013/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0013/README.md
@@ -6,20 +6,18 @@ A search path, such as `$PATH` in `bash`, is a list of directories used to find 
 
 Python has a 'hard coded' and 'soft coded' search path lists:
 
-    * `sysconfig` determined during installation 
-    * `sys.path` determined at runtime using:
-    ** Script/current dir
-    ** Site-packages
-    ** Standard library (`/usr/lib/python3.xx`)
-    ** Environment variable `PYTHONPATH` defined by the operating system or runtime shell
+* `sysconfig` determined during installation
+* `sys.path` determined at runtime using:
+  * Script/current dir
+  * Site-packages
+  * Standard library (`/usr/lib/python3.xx`)
+  * Environment variable `PYTHONPATH` defined by the operating system or runtime shell
 
 The `example01.py` code prints the content of `sysconfig` and `sys.path`.
 
-## Example Code Example
-
 _[example01.py](example01.py)_:
 
-```python
+```py
 # SPDX-FileCopyrightText: OpenSSF project contributors
 # SPDX-License-Identifier: MIT
 """Example Code Example"""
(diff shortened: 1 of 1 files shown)


docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0013/README.md
```diff
@@ -1,7 +1,9 @@
-# pyscg-0013: Untrusted Search Path
+# pyscg-0013: Secure Search Paths
 
 In an environment where an untrusted or less trusted entity can modify the environment variables, consider validating hash-based byte code [Python 2023 Command line and environment].
 
+The search path module is a list of directories that are searched when importing modules and packages using `import`. This search path is storeed in `sys.path` [[note.nkmk.me 2023]](https://note.nkmk.me/en/python-import-module-search-path/). An attacker who can modify these paths may inject malicious modules that get loaded instead of legitimate ones.
```
myteron: hm might be a bit long winded but turns out that its actually less streight forward then what I thought. 
If the inline code is to messy then we have to stick it in an example code. 

```suggestion
A search path, such as `$PATH` in `bash`, is a list of directories used to find scripts, modules or binaries.

Python has a 'hard coded' and 'soft coded' search path lists:

 * `sysconfig` determined during installation 
 * `sys.path` determined at runtime using:
 ** Script/current dir
 ** Site-packages
 ** Standard library (`/usr/lib/python3.xx`)
 ** Environment variable `PYTHONPATH` defined by the operating system or runtime shell

The `example01.py` code prints the content of `sysconfig` and `sys.path`.

# SPDX-FileCopyrightText: OpenSSF project contributors
# SPDX-License-Identifier: MIT
"""Example Code Example"""

import sysconfig
import sys

print("sysconfig.get_paths().values())", end=" ")
print(sysconfig.get_paths().values())
print("")
print("print(sys.path)", end=" ")
print(print(sys.path))

An attacker who can modify a search paths via environmental variables or other means can potentially inject malicious modules that get loaded instead of legitimate ones.

Consider validating hash-based byte code [[Python 2023 Command line and environment](https://docs.python.org/3.9/using/cmdline.html#cmdoption-check-hash-based-pycs)].
```

dwiley258: If the original is used stored is spelt incorrectly.
Other than this everything else looks good from my side.

BartKaras1128: I think that's updated now Helge, have a look and let me know

BartKaras1128: We did not use the original, I think it should be OK now. Have a look and let me know!

PR #1049, opened 2026-02-23
Updates GitHub Pages templates to add theme, header and footer

GeauxJD:
Updates to the best.openssf.org GitHub pages, to add OpenSSF header and footer content as we have done in other repositories like policy.openssf.org and baseline.openssf.org.

Here are screenshots of the header and footer from testing these updates in my local environment:

Screenshot 2026-02-23 at 6 02 24 PM
Screenshot 2026-02-23 at 6 02 34 PM


gkunz (approved):
Thanks!

PR #1067, opened 2026-03-11
Explicitly mention improved memory safety as benefit in Compiler Annotations Guide

thomasnyman:
Based on past discussion with @david-a-wheeler


thomasnyman:
Merging as per discussion with @gkunz

PR #1072, opened 2026-03-13
Add draft AI/ML Driven SW Development guide

david-a-wheeler:
We already have a course, but this ensures that people can *also* get this as a guide.

It's also an example of a guide we can create and later turn into a course (the other direction), which gives a starting example for the next course we hope to create.


mihaimaruseac (approved):
This looks great! Thank you!

gkunz (approved):
We agreed to merge this as a draft to get it into the repo. Work will continue on the slides as the single source of truth.

PR #1083, opened 2026-04-07
pyscg-0037: Tweak heading test about assertions

david-a-wheeler:
It's not that you *can't* use assertions, it's that you need to not depend on their utility in production. Leaving them in is fine, as long as you don't depend on them firing during production.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
 # pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are removed in production.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are disabled in production.
 
 Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
 # pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment. 
+Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment.
 Only use `assert` for internal invariants and debug-only checks that are *not required for correct or safe execution*.
 
 __Do not use `assert` for:__
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
 # pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are disabled in production.
+Assertions are a useful developer tool, but they cannot be relied upon to be enabled in a production environment. Only use assertions when disabling them in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of using assertions, to ensure that they will continue to occur even when assertions are disabled in production.
 
 Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
 # pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Assertions can be used, but incorrect function arguments should be handled by an appropriate exception instead.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are removed in production.
 
 Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,4 +1,4 @@
-# pyscg-0037: Do not Depend on Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
 
 Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Assertions can be used, but incorrect function arguments should be handled by an appropriate exception instead.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -12,7 +12,7 @@ __Reason:__
 
 Python’s optimization options (`-O` and `-OO`) remove `assert` statements from the bytecode and set `__debug__` to `False`  [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. This means any logic implemented using `assert` may not execute in optimized runs.
 
-As a result, using `assert` for security or correctness-critical checks can introduce vulnerabilities or undefined behavior regardless if optimization is enabled. Its neither recommended to depend on `-O` or `-OO` as a "switch" to make assertions unreachable. 
+As a result, using `assert` incorrectly (that is, for security or correctness-critical checks) can introduce vulnerabilities or other undesired behavior when optimization is enabled.
 
 Instead, raise appropriate exceptions to ensure the checks are always enforced regardless of interpreter settings.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Do not Depend on Assertions In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Assertions can be used, but incorrect function arguments should be handled by an appropriate exception instead.
 
 Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
 
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
@@ -1,6 +1,6 @@
 # pyscg-0037: Presume Assertions May Be Disabled In Production
 
-suggestion
+Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment. 
 Only use `assert` for internal invariants and debug-only checks that are *not required for correct or safe execution*.
 
 __Do not use `assert` for:__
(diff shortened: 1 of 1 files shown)


docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Do not Depend on Assertions In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Assertions can be used, but incorrect function arguments should be handled by an appropriate exception instead.
```
myteron: @david-a-wheeler 

I believe that basically 'allowing' assertion code to be present in production code requires to change both sentences and be very clear on when it is safe to have assertions present including reference's supporting that statement.

myteron:
Moving discussion in #1085 here.

@david-a-wheeler please note that MITRE had advised us to change all titles to "do's" instead of "don'ts". While this is not always possible we did spend 3-4 month to refactor the titles as definition in [CONTRIBUTING.md#rule-titles](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/CONTRIBUTING.md#rule-titles)

> Titles should:
> *  Be written in APA Title Case style [[APA 2022](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case)]
> *  Be action-oriented
> *  Preferably describe wanted design over anti-pattern

**Alternative title options:**

* Do not Depend on Assertions In Production
* Avoid Reachable Assertions
* Only depend on assertions outside of production.

Obviously we end up having to explain exactly and what type of assertions are ok to have code remaining in production i.e. "not reachable".

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
david-a-wheeler: Forbidding the use of assertions means you have to re-create them, which would be worse. Fuzzers are *much* more effective, for example, when you have assertions, because assertions make it possible for fuzzers to detect more problems.

You *want* assertions in the code, because they make testing more productive. However, since assertions are often disabled in production Python code, the code must work properly when assertions are removed.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Do not Depend on Assertions In Production
```
david-a-wheeler:
```suggestion
# pyscg-0037: Presume Assertions May Be Disabled In Production
```

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
david-a-wheeler:
```suggestion
Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are removed in production.
```

BartKaras1128: Hi @david-a-wheeler.
I wonder if we have different meanings of the word "production".

Our understanding was something like this:
Dev = your laptop while coding
Staging = a private preview before launch
Production = the public website everyone visits

We would not say that unit tests are expected to be in a well designed production environment. Perhaps if there's different interpretations of production, we have to do a better job clarifying this in the rule.

I would argue that we are not completely restricting the use of assertions in the rule, as we are saying "Avoid" which basically says "consider something else".

> since assertions are often disabled in production Python code, the code must work properly when assertions are removed.
 
Our understanding is that test code is not supposed to be in production code / live systems.

I would also argue that even if there is usecases in which we want assertions to be in production code, it would be contradictory to what is in the rule at the moment as optimising it via -O/-OO would completely break the code.

Also, generally speaking using -O/-OO to switch between staging and live production is generally a bad idea.

Don't mean to be a pain about this, apologies if that's how it's coming across that way!

s19110: @david-a-wheeler @BartKaras1128 

Personally, I have not come across any uses of assertions outside of unit tests, so I did my best to check the literature on the topic.

Perhaps we should use the term "public release" (or just "release") instead of "production" to avoid confusion.

"Programming in Python 3" (ISBN-13: 978-0-321-68056-3, p.184-185) does support the idea of optimising the execution via the -O/-OO flags, or even creating a separate production code with the assertions commented out.

"Clean Code in Python" (ISBN 978-1-80056-021-5, ch. "Using assertions in Python", p. 88-90) stresses assertions must not be used when evaluating business logic, but it does suggest they can be used to perform sanity checks and verifying invariants to make sure our algorithms themselves are correct. It even includes the following tip, which discourages from using the -O flag in production:

> Do not run your production programs with python –O … as 
you'd like to take advantage of the assertions in the code to correct 
defects.

It is difficult to find the consensus on the matter but it appears as though assertions in production code are permissable as long as optimisation isn't essential and they aren't used to evaluate business logic. If we want to prevent a scenario where an assumed impossible condition occurs and we have no way of recovery, even a crash using assertions might be an acceptable solution.

david-a-wheeler:
Assertions are really valuable when running tests and fuzzers. Tests and fuzzers cannot guarantee to detect all defects. However, the likelihood of their detecting defect increases when you can increase the number of problems they can detect. Adding assertions can dramaticaly improve their effectiveness.

However, it's inappropriate to "add assertion lines during testing/fuzzing and remove the lines before shipping". The whole *REASON* basically every language has an assert statement is to enable easy control on whether or not to run them. You want lots of assert statements in code, you just don't want to have the business or security logic *depend* on assertions being enabled (because that is not guaranteed).

david-a-wheeler:
I made some tweaks. Is this clearer?

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are removed in production.
```
david-a-wheeler:
```suggestion
Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are disabled in production.
```

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,6 +1,6 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Only use assertions when their removal in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of assertions, to ensure that they will continue to occur even when assertions are disabled in production.
```
david-a-wheeler:
```suggestion
Assertions are a useful developer tool, but they cannot be relied upon to be enabled in a production environment. Only use assertions when disabling them in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of using assertions, to ensure that they will continue to occur even when assertions are disabled in production.
```

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,6 +1,6 @@
...
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be enabled in a production environment. Only use assertions when disabling them in production is acceptable. In particular, to report invalid input or incorrect function arguments, raise an appropriate exception or other such actions instead of using assertions, to ensure that they will continue to occur even when assertions are disabled in production.
 
 Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
```
myteron: @david-a-wheeler after closer inspection of assert I believe now that you are correct about assert statements. 

A blanket prohibition of `assert` is not appropriate and would lead to false positives, as it is legitimately used in the Python standard library.
see `egrep -Ru assert /usr/lib/python3.9/ 2>/dev/null|egrep -v "unittest|test_support|test"`.

The issue I see now is that its not about *where* but *how* `assert` is used. It looks like the discussion got a bit derailed, apologies for that. 

```suggestion
suggestion
Only use `assert` for internal invariants and debug-only checks that are *not required for correct or safe execution*.

__Do not use `assert` for:__
* Security checks (including input validation)
* Enforcing runtime conditions whose removal would affect correctness or security
* Error handling

__Reason:__

Python’s optimization options (`-O` and `-OO`) remove `assert` statements from the bytecode and set `__debug__` to `False`  [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. This means any logic implemented using `assert` may not execute in optimized runs.

As a result, using `assert` for security or correctness-critical checks can introduce vulnerabilities or undefined behavior regardless if optimization is enabled. Its neither recommended to depend on `-O` or `-OO` as a "switch" to make assertions unreachable. 

Instead, raise appropriate exceptions to ensure the checks are always enforced regardless of interpreter settings.
```

Current title is "pyscg-0037: Avoid Assertions In Production"

We have to remove 'production environment' from our current title, suggested new title: 

@david-a-wheeler : We have been discussing the titles today, came up with more variations and voted:

A) Restrict assert to Internal Non-Critical Checks 3/5
counter argument here was that "Non-Critical Checks" sounds to vague. Also one would only have asserts for use-full and required internal invariant checks. 
 
B) Do not Depend on Assertions in Production 3/5
Counter argument, still holds onto "production".

C) Use Assertions Only for Internal Invariants and Debug-Only Checks 5/5
While it does miss out on "reachable" it does capture the only reasons for us to leave assert in production code. The title would get to long tho, so capping it to only the essence. 

Rg Helge

myteron (commented):
See suggested doc change.

myteron (changes requested):

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,20 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+suggestion
```
david-a-wheeler:
```suggestion
Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment. 
```

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,20 @@
...
+__Do not use `assert` for:__
+* Security checks (including input validation)
+* Enforcing runtime conditions whose removal would affect correctness or security
+* Error handling
+
+__Reason:__
+
+Python’s optimization options (`-O` and `-OO`) remove `assert` statements from the bytecode and set `__debug__` to `False`  [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. This means any logic implemented using `assert` may not execute in optimized runs.
+
+As a result, using `assert` for security or correctness-critical checks can introduce vulnerabilities or undefined behavior regardless if optimization is enabled. Its neither recommended to depend on `-O` or `-OO` as a "switch" to make assertions unreachable.
```
david-a-wheeler:
```suggestion
As a result, using `assert` incorrectly (that is, for security or correctness-critical checks) can introduce vulnerabilities or other undesired behavior when optimization is enabled.
```

It won't be *undefined* behavior in Python, that's a C-ism.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,20 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment.
```
david-a-wheeler:
```suggestion
Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment.
```

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,20 @@
...
+Only use `assert` for internal invariants and debug-only checks that are *not required for correct or safe execution*.
 
-Python removes assertions when a script is run with the `-O`  and `-OO` options [[Python 3.9 Documentation](https://docs.python.org/3.9/using/cmdline.html?highlight=pythonoptimize#cmdoption-o)]. The `-O` options is for optimisation. It removes asserts statements from bytecode, removes docstrings from functions/classes and sets `__debug__` to False. It is used for slightly faster execution and smaller bytecode files. `-OO` does everything that `-O`does but it additionally removes module-level docstrings and creates an even more compact bytecode.
+__Do not use `assert` for:__
+* Security checks (including input validation)
```
david-a-wheeler, line 8:
```suggestion

* Security checks (including input validation)
```

david-a-wheeler:
We have the same kind of definition of Production, e.g., "real world use" such as "the public website everyone visits". Assertions don't necessarily have anything to do with unit tests.

The title `Presume Assertions May Be Disabled In Production` would be a good title, by the way. That is a positive statement.

> Personally, I have not come across any uses of assertions outside of unit tests, so I did my best to check the literature on the topic.

My experience is different. I've seen many programs with many assertions in the production code, and they can be *really* helpful for detecting defects before production use. They're not just for unit tests.

Let's imagine you're doing integration tests, fuzzing, and so on. Assertions are left enabled. In the *production* code (*NOT* the test code) are many assertions like this:

~~~python
    # ... do complicated things ...
    assert i < j
    # ... more complicated things ...
~~~

Now, if the integration test, fuzzer, etc., triggers a case where `i < j` is false, the assertion *immediately* triggers. If you have many assertions, your integration test, fuzzing, etc., are *far* more effective, because not only are you checking the end result or "not crash", you're constantly peering into the system's internal state to ensure it's sensible. Not only are you more likely to *detect* a fault, but *fixing* is much faster because you have earlier information on something that went wrong.

The more you sprinkle your production code with useful assert statements, the more effective your test process becomes, but only if the assert statements *stay* in the production code. Disabling them with -O0 is fine and expected. That's why the flag exists! Trying to delete & re-insert assert statements in production code is a terrible idea. The whole *reason* for the optimization flag is that you can leave the assertions *in* the production code, and disable their execution whenever you want to.

Anyway, I hope all this discussion helps!

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
david-a-wheeler: > it appears as though assertions in production code are permissable as long as optimisation isn't essential and they aren't used to evaluate business logic

Exactly. Assertions within production code are a powerful and useful technique, which is why practically all programming languages include them. But like all constructs, they have to be used *correctly* - if you use asserts or any other construct incorrectly, it's a bad idea.

myteron (approved):

s19110 (approved):
The tweaked version of the rule sounds good. @david-a-wheeler, thank you for the pull request and the productive discussion.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,21 @@
-# pyscg-0037: Avoid Assertions In Production
+# pyscg-0037: Presume Assertions May Be Disabled In Production
```
tommcd, line 1: If we are following APA style then 'in' should be lowercase:
https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
```diff
@@ -1,8 +1,21 @@
...
 
-Assertions are a useful developer tool, but they cannot be relied upon to be present in a production environment. Incorrect function arguments should be handled by an appropriate exception.
+Assertions are a useful developer tool, but they cannot be relied upon to be applied in a production environment.
```
tommcd, line 3: This sentence is slightly off:

they cannot be relied upon to be applied in a production environment.

“Applied” is not quite the right verb. The Python docs are more precise: with optimization, __debug__ becomes False and the code generator emits no code for assert. So the issue is that assertions may not execute at all, not that they are not “applied.”

I would rewrite it as:

Assertions are a useful developer tool, but production code must not rely on them executing.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0037/README.md
tommcd: I wish we weren't following APA style, but I lost that argument :-)

PR #1090, opened 2026-04-19
PSCG: fixing links in html tables

gkunz:
This removes the target markdown file from <a> links used in html tables as these do not get rendered correctly on GitHub pages and thus point to the plain text markdown files on best.openssf.org

Fixes #1089

PR #1092, opened 2026-04-23
adding a copy of the readme with relative links for the web presentation

myteron:
short term solution to allow the web links to work without breaking links in a coding ide as an answer to #1090


gkunz (approved):
There are unfortunately only sub-optimal solutions for this problem. Let's go with this for the time being and try to keep the content consistent (the biggest issue I see with this approach)

BartKaras1128 (approved):
Looks good to me!

PR #1101, opened 2026-05-20
Update the deserialization section

david-a-wheeler:
The previous text was simple but often wrong or omitting key practical guidance. This text strongly recommends JSON (one of the few simple clearly-safe ways), then explains how to handle the many cases where people won't do that.

This notes that XML is safe, but only in some circumstances, and exaplaining how to determine if it's safe.
For example, on my system loading XML in Python is *NOT* *SAFE*, but this is something that can be programmatically determined. Also, you should consider that XML is not safe to load by default in other systems, and thus you might consider avoiding it if there are reasonable alternatives.

Similarly, YAML's `load` is *NOT* safe, you have to use `safe_load`.

Saying "don't use pickle" is true but unhelpful advice. ML is the big rage, and many models are in pickle format. So are many other data values that you need.
So yes, recommend against using pickle if at all possible. However, since people *MUST* load this format sometimes, it's important to show the painful process necessary to securely load pickle when *do* have to do it. Showing what you have to do is a better argument against pickle :-).

I wish this were simpler. However, I think it's more important to be accurate, and to explain what you have to do.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0023/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0023/README.md
@@ -1,20 +1,31 @@
-# pyscg-0023: Secure Deserialization
+# pyscg-0023: Secure Deserialization (including pickle)
 
-Even if data has been created from a trusted source, we need to verify that it has not been tampered with during transport.
+When loading data into your program ("deserializing"), do not permit an attack. Only load data (1) from a trusted source *or* (2) by using a secure loading process. Where practical, do both.
 
-The `pickle` module is known to be vulnerable [[docs.python.org 2023]](https://docs.python.org/3.9/library/pickle.html) against unwanted code execution during deserialization and should only be used if there is no architectural text-based alternative.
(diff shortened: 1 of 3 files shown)


david-a-wheeler:
I imagine we'll have discussions on how to simplify this text, but I think the current text overcorrects, focusing on making it "simple but wrong". I think it's more important to be correct, and then as simple as we can while *staying* correct.

myteron:
Hi @david-a-wheeler , the current text is trying to following advise by the official python documentation as of 3.14.6 https://docs.python.org/3/library/pickle.html. Specifically as of 
>Warning
>The pickle module is not secure. Only unpickle data you trust.
> It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.
> Consider signing data with [hmac](https://docs.python.org/3/library/hmac.html#module-hmac) if you need to ensure that it has not been tampered with.
> **Safer serialization formats such as [json](https://docs.python.org/3/library/json.html#module-json)** may be more appropriate if you are processing untrusted data. See [Comparison with json](https://docs.python.org/3/library/pickle.html#comparison-with-json).

As much as I appreciate  a solution to make usage Pickle more security, it is not the intent on pyscg an instruction set of how to make unsecure modules fully secure. Aside from generally overloading the document with documentation and code it would result in a moving target, basically becoming a framework. This is not our intent.

We are totally happy to link in external resources providing such details but have to keep in mind the total amount of hours for study or any probability of a human to remember its content. Any quantity of external content with details on any if this is no issue at all. 

Rg Helge

PR #1103, opened 2026-06-09
Add length parameter to os.read

HoBoIs:
Pythons os.read requires 2 parameters, but only 1 was given in the test file making it a test for missing parameters, what is not intended.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0015/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0015/README.md
@@ -35,7 +35,7 @@ import uuid
 def read_file(file):
     """Function for opening a file and reading it's content."""
     fd = os.open(file, os.O_RDONLY)
-    content = os.read(fd)
+    content = os.read(fd, 1024)
     return content.decode()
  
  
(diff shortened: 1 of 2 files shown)


06kellyjac (commented):
looks good but needs [DCO signoff](https://wiki.linuxfoundation.org/dco)

HoBoIs:
> looks good but needs [DCO signoff](https://wiki.linuxfoundation.org/dco)

Done

HoBoIs:
@06kellyjac gentle ping.

HoBoIs:
@06kellyjac  or any other developer, could you take a look and merge if it is OK? I fixed the DCO signoff, and it should be fine now.

gkunz:
Hi @HoBoIs. Thank you very much for your contribution and sorry for the delayed response. I have add added @myteron and @BartyBoi1128 to this PR as reviewers. They are the maintainers of this guide.

myteron:
@HoBoIs apologies for getting back on you. Downside of existing in a big repo with tons of other stuff is that we don't see this unless its having the label.

BartKaras1128 (approved):
Looks good @HoBoIs!

Apologies for the reviewing delay, this never had the "Product: Python Hardening Guide" tag on it so I never saw it!

On my version of Python3.13.2 it never actually reached that line so the error raised was the exact same for me in both cases, however that doesn't change the fact that 2 parameters should have been included in os.read!

Great spot!

myteron:
2x+1, merging. Thanks @HoBoIs !

PR #1106, opened 2026-06-30
Bump date for Compiler Hardening Guides

thomasnyman:
2026-06-30


thomasnyman:
Fix trailing character in DCO and use correct author e-mail

Issue #992, opened 2025-09-19
pySCG-0042: make code examples clearly demonstrate unintended side effects when failing to Ensure Correct Operator Precedence,


myteron:
Need to come with better code examples as per comments below:

The “Non-Compliant” snippet uses pure expressions (bit ops, tuple permutation) with no side effects. The introduction claims risk from “expressions that read & write to the same object”.

The compliant solution has identical output. While it's certainly much easier to understand and maintain,
given the introduction I would expect examples that demonstrate "unintended side effects" or a result that is "surprising and incorrect"

BartKaras1128:
I will look into this now!

BartKaras1128:
This is merged now so issue can be closed @myteron

PR #1108, opened 2026-07-06
issue-992: pySCG-0042: make code examples clearly demonstrate uninten…

BartKaras1128:
…ded side effects when failing to Ensure Correct Operator Precedence.

A pull request addressing this issue: https://github.com/ossf/wg-best-practices-os-developers/issues/992

Code examples are taken from real world CVE that I found here, obviously oversimplified for educational purposes:
https://nvd.nist.gov/vuln/detail/CVE-2026-7270

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0042/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0042/README.md
@@ -51,95 +51,72 @@ If a method changes an object’s state (has side effects) and is called multipl
 
 ## Non-Compliant Code Example
 
-`noncompliant01.py` is expected to provide labels for numbers, but it unnecessarily obfuscates the evaluation and logic.
+`noncompliant01.py` demonstrates an operator precedence logic error that can lead to a buffer overflow vulnerability. This code example is based on a real vulnerability ([CVE-2026-7270](https://nvd.nist.gov/vuln/detail/CVE-2026-7270)) that occurred in 'FreeBSD's' execve argument handling. The code fails to use parentheses to clarify the intended order of operations, causing the bounds check to pass when it should fail.
 
 _[noncompliant01.py](noncompliant01.py):_
 
 ```python
+# SPDX-FileCopyrightText: OpenSSF project contributors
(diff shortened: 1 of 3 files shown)


s19110:
reviewing

docs/Secure-Coding-Guide-for-Python/01_introduction/pyscg-0042/README.md
```diff
@@ -51,96 +51,73 @@ If a method changes an object’s state (has side effects) and is called multipl
...
+Bounds check passed: remaining=15
 ```
 
-The `noncompliant01.py` does respond with the correct output. Extending the `noncompliant01.py` to also a label `postive` or `zero` numbers would be challenging.
+The bounds check incorrectly passes because the expression evaluates left-to-right as `(16 - 10) + 9 = 15` instead of the intended `16 - (10 + 9) = -3`. This allows the attacker-controlled data to overflow the buffer.
 
 ## Compliant Solution
 
-This compliant solution, uses equivalent logic and performs at most one write operation per expression, which makes the code easier to understand and maintain.
+The compliant solution fixes the operator precedence by comparing additions instead of subtracting, making the code clear, with no precedence trap. While adding parentheses like `ARG_MAX_BYTES - (USED + LENGTH)` would fix the precedence issue also, restructuring to `USED + LENGTH > ARG_MAX_BYTES` is clearer and avoids confusing negative values.
```
s19110, line 89: Nitpick: I found the wording a bit confusing. How about this?

```suggestion
The compliant solution avoids the operator precedence trap by comparing additions instead of subtracting, making the code clearer. While adding parentheses like `ARG_MAX_BYTES - (USED + LENGTH)` would also fix the precedence issue, restructuring to `USED + LENGTH > ARG_MAX_BYTES` is more readable and avoids confusion caused by negative values.
```

s19110 (approved):
Looks good to me! I like this code example much better than the original because it's much simpler while still potraying the core issue well. I added one minor suggestion regarding wording.

dwiley258 (approved):
Looks good to me

myteron (approved):
all changes applied

PR #1114, opened 2026-07-14
Add -fzero-call-used-regs to the C/C++ Compiler Options Hardening Guide

### LINKED ISSUE #1110: Consider `-fzero-call-used-regs` for C and C++ Compiler Hardening Guide
`-fzero-call-used-regs` is a compiler hardening flag in GCC and Clang that automatically overwrites ("scrubs") hardware registers with zeros before a function returns. It limits the lifetime of sensitive data, preventing information leaks and reducing available gadgets for Return-Oriented Programming (ROP) attacks.

The specified value of the options control the set of registers that are subject to scrubbing:
- `skip`: Don't zero out any registers. This is the default.
- `used`: Zero out all used registers.
- `used-arg`: Zero out used registers that are used for arguments.
- `used-gpr`: Zero out used registers that are GPRs.
- `used-gpr-arg`: Zero out used GPRs that are used as arguments.
- `all`: Zero out all registers.
- `all-arg`: Zero out all registers used for arguments.
- `all-gpr`: Zero out all GPRs.
- `all-gpr-arg`: Zero out all GPRs used for arguments.

Resources:
- https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fzero-call-used-regs
- https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fzero-call-used-regs
- https://blog.quarkslab.com/clang-hardening-cheat-sheet-ten-years-later.html

skypank-coder:
# Add `-fzero-call-used-regs` to the C/C++ Compiler Options Hardening Guide

Closes #1110.

## Placement decision: RECOMMENDED (Table 2), not the considered/rejected appendix

The guide's appendix is where flags go to die, and two of its residents are the
obvious comparison points. This flag does not resemble either.

**Against `-mmitigate-rop`** (appendix: "considered to be ineffective and
security-theatre-esque, so was deprecated in GCC 9.1"). That flag tried to avoid
emitting opcodes reinterpretable as `ret`, was never adopted, and upstream GCC removed
it. `-fzero-call-used-regs` is the opposite on every axis: it is actively maintained and
still gaining values (`leafy` in GCC 14), it is adopted upstream by the Linux kernel
(`CONFIG_ZERO_CALL_USED_REGS`, since v5.15) and OpenSSH, and its effect is *measured*
rather than asserted — the kernel reports ~20% fewer unique ROP gadgets and the
elimination of compiler-generated "write-what-where" gadgets, and an independent
measurement reports 36.5% and notes that automatic ROP chain generation fails in its
early stages for want of GPR write gadgets. Deprecated-and-ineffective versus
adopted-and-measured is the whole distinction.

**Against `-mshstk`** (appendix: redundant, since "CET instrumentation is controlled by
`-fcf-protection`"). `-mshstk` is rejected because it is the *same mechanism* as a flag
we already recommend. `-fzero-call-used-regs` is a *different layer*: CET protects the
control flow itself (shadow stack validates the return address; IBT constrains indirect
branch targets), whereas register zeroing performs no control-flow check at all — it
shortens the lifetime of register data and starves gadget chains of the register writes
that make them useful. They mitigate different stages of the same attack. They are also
complementary in a sharper way: register zeroing *increases* JOP gadgets ~13%, and JOP
is precisely the class IBT constrains. Recommending both is coherent; recommending
`-mshstk` alongside `-fcf-protection` would not be. This argument is stated explicitly
in **Additional Considerations** so the "isn't this redundant with CET?" question is
answered in the guide, not just in this PR.

The honest case against Table 2 is that the mitigation is circumventable (misaligned
offsets still yield usable gadgets) and that it *adds* JOP gadgets. Both facts are in
the section, framed as defense in depth rather than a silver bullet. Neither is
disqualifying: by that standard `-fstack-protector-strong` (bypassable) and
`-fcf-protection` (incomplete) would also fail. The cost is bounded and documented
(<1% perf), and the flag is a compile-and-go option requiring no source changes — which
is the bar Table 2 sets.

## Changes

1. **One row** in Table 2 (Recommended compiler options that enable run-time protection
   mechanisms), placed after `-mbranch-protection=standard` with the other ROP/JOP
   mitigations.
2. **One new `###` section** placed directly after the `-fcf-protection` /
   `-mbranch-protection` section, mirroring its structure: variant table with
   `<span id="…">` anchors, `#### Synopsis`, `#### Performance implications`,
   `#### When not to use?`, `#### Additional Considerations`.
3. **Seven footnotes**, all to primary sources, defined inline in the existing style.

No other content is touched: 57 insertions, 0 deletions. The scraper is unchanged — it
reads Table 2 automatically, and its version regex intentionally caps the final version
node at one digit so footnote superscripts are not absorbed; the new row parses as
`{"gcc": "11.0.0", "clang": "16.0.0"}`. `markdownlint` reports 0 errors, matching the
pre-change baseline.

## Verified facts and sources

Every version number and claim below was confirmed against a primary source; nothing is
from memory.

| Claim | Source |
|---|---|
| Flag exists in GCC 11; documented under **Optimize Options** (not Instrumentation) | [GCC Manual, Optimize Options](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fzero-call-used-regs); GCC 11.1.0 manual confirms presence |
| Full GCC value taxonomy (`skip`/`used`/`all` × `-gpr`/`-arg`), and the `zero_call_used_regs` function attribute | [GCC Manual, Common Function Attributes](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) |
| GCC **14** adds the `leafy` family (absent from the GCC 11/12/13 manuals) | Diffed the GCC 11.1.0 / 12.1.0 / 13.1.0 / 14.1.0 attribute manuals |
| Clang values: `skip`, `used-gpr-arg`, `used-gpr`, `used-arg`, `used`, `all-gpr-arg`, `all-gpr`, `all-arg`, `all` — **no `leafy`**; targets **AArch64 / RISC-V / x86 only** | [Clang Command Line Argument Reference](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fzero-call-used-regs) (generated from `Options.td`) |
| Clang introduced the option in **15.0.0** | [Clang 15.0.0 Release Notes](https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html) |
| Clang 15 miscompiles → NULL pointer dereferences | [llvm/llvm-project#57692](https://github.com/llvm/llvm-project/issues/57692) |
| Kernel gates on `clang > 15.0.6`; fix ships in **Clang 16.0.0** | Linux commit [`d6a9fb87e9d1`](https://git.kernel.org/linus/d6a9fb87e9d1) (Chancellor, 2022-12-15) |
| That gate was removed in 2026 only because the kernel's minimum LLVM rose to 17.0.1 — i.e. it lapsed, it was not repudiated | Linux commit [`813fe686e90b`](https://git.kernel.org/linus/813fe686e90b) (2026-05-27) |
| Kernel builds with `used-gpr`; ~20% fewer unique ROP gadgets; removes "write-what-where" gadgets; **<1% perf**; image growth <1% x86_64 / ~5% arm64 | Linux commit [`a82adfd5c7cb`](https://git.kernel.org/linus/a82adfd5c7cb) (Cook, 2021-07-20) + `security/Kconfig.hardening` + top-level `Makefile` |
| Independent measurement: **36.5%** unique ROP gadget reduction, **+13%** JOP/SYS gadgets, circumventable via misaligned offsets, ROP chain generation fails early | Björklund, Jerker, [Gadget reduction using zero-call-user-regs](https://www.jerkeby.se/newsletter/posts/rop-reduction-zero-call-user-regs/), 2021-11-26 |
| OpenSSH uses `-fzero-call-used-regs=used`; disables it on Clang 17 (ICE) | [openssh-portable `configure.ac`](https://github.com/openssh/openssh-portable/blob/master/configure.ac); [llvm/llvm-project#69794](https://github.com/llvm/llvm-project/issues/69794) |

Two source-conflict notes, resolved in favour of transparency rather than picking a
number: (a) the kernel says **~20%** ROP gadget reduction while the independent
measurement says **36.5%** — the section cites both, attributed. (b) The `+13%` JOP
increase is reported only by the independent measurement; it is stated anyway, because
omitting it would misrepresent the flag.

## Question for maintainers (deliberately not decided here)

Table 2 options are generally mirrored in the **TL;DR** block, where this flag would
belong on the `for x86_64` line. I did not add it, to keep this diff reviewable and
because the target story is not purely x86_64 (GCC's kernel usage covers arm64; Clang
supports AArch64/RISC-V/x86). Happy to add it in a follow-up commit if you want TL;DR
parity, and to take direction on how to phrase the architecture qualifier.


skypank-coder:
@thomasnyman Please review this .

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
 
 [^gcc-release-notes-14]: GCC team, [GCC 14 Release Series Changes, New Features, and Fixes](https://gcc.gnu.org/gcc-14/changes.html), 2024-08-10.
 
+### Zero call-used registers on function return to limit register data lifetime and ROP gadgets
+
+| Compiler Flag                                                                     | Supported since             | Description                                                                                                       |
+|:----------------------------------------------------------------------------------|:---------------------------:|:------------------------------------------------------------------------------------------------------------------|
+| <span id="-fzero-call-used-regs=used-gpr">`-fzero-call-used-regs=used-gpr`</span> | GCC 11.0.0<br/>Clang 16.0.0 | Zero call-used general-purpose registers that the function used, on return.                                       |
```
thomasnyman: The "call-used" in the description makes the descriptions awkward to read, especially when the "`all`*" variants of the options zero out all, not used call-used registers.

It's also unclear which function "the function" refers to her since the option applies to all functions.

```suggestion
| <span id="-fzero-call-used-regs=used-gpr">`-fzero-call-used-regs=used-gpr`</span> | GCC 11.0.0<br/>Clang 16.0.0 | Zero out general-purpose registers used by functions on return.                                       |
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+| <span id="-fzero-call-used-regs=used">`-fzero-call-used-regs=used`</span>         | GCC 11.0.0<br/>Clang 16.0.0 | Zero all call-used registers that the function used, on return, including registers that are not general-purpose. |
```
thomasnyman:
```suggestion
| <span id="-fzero-call-used-regs=used">`-fzero-call-used-regs=used`</span>         | GCC 11.0.0<br/>Clang 16.0.0 | Zero out all registers that the function used on return, including registers that are not general-purpose. |
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+| <span id="-fzero-call-used-regs=all-gpr">`-fzero-call-used-regs=all-gpr`</span>   | GCC 11.0.0<br/>Clang 16.0.0 | Zero all call-used general-purpose registers on return, whether the function used them or not.                    |
```
thomasnyman: Either "general-purpose" registers or "all registers" to be consistent with the wording of descriptions for all variants; "all general-purpose"  breaks the pattern.

```suggestion
| <span id="-fzero-call-used-regs=all-gpr">`-fzero-call-used-regs=all-gpr`</span>   | GCC 11.0.0<br/>Clang 16.0.0 | Zero out general-purpose registers on return, whether the function used them or not.                    |
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+| <span id="-fzero-call-used-regs=all">`-fzero-call-used-regs=all`</span>           | GCC 11.0.0<br/>Clang 16.0.0 | Zero all call-used registers on return, whether the function used them or not.                                    |
```
thomasnyman:
```suggestion
| <span id="-fzero-call-used-regs=all">`-fzero-call-used-regs=all`</span>           | GCC 11.0.0<br/>Clang 16.0.0 | Zero out all registers on return, whether the function used them or not.                                    |
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+| <span id="-fzero-call-used-regs=skip">`-fzero-call-used-regs=skip`</span>         | GCC 11.0.0<br/>Clang 16.0.0 | Zero no call-used registers (the default).                                                                        |
```
thomasnyman: More direct language:

```suggestion
| <span id="-fzero-call-used-regs=skip">`-fzero-call-used-regs=skip`</span>         | GCC 11.0.0<br/>Clang 16.0.0 | Do not zero out any registers (the default).                                                                        |
```

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+
+#### Performance implications
+
+The cost is proportional to the number of registers zeroed on each function return, and so increases across `used-gpr`, `used`, `all-gpr`, and `all`; the `all` forms pay to zero registers the function never touched. For `used-gpr`, the Linux kernel documents a performance impact of less than 1% on most workloads[^kernel-zero-call-used-regs].
+
+Binary size also grows, and the growth is architecture-dependent: the kernel reports image growth of less than 1% on x86_64 and about 5% on arm64[^kernel-zero-call-used-regs]. Projects that are sensitive to binary size should measure the impact on their own target before enabling the option.
+
+#### When not to use?
+
+Do not use this option with versions of Clang older than 16.0.0. Clang implemented `-fzero-call-used-regs` in 15.0.0, but a code generation bug in that implementation could produce invalid code resulting in NULL pointer dereferences at run time[^clang-zero-call-used-regs-bug].
```
thomasnyman, line 916: It seems somewhat contradictory to recomment the option in 16.0.0 but then point to potential bugs in the feature present in Clang 17 ([[^openssh-zero-call-used-regs]](https://github.com/llvm/llvm-project/issues/69794)). Was there any conclusion as to whether the regression in Clang no longer affects newer versions of Clang?

docs/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.md
```diff
@@ -878,6 +879,62 @@ AArch64 BTI and PAC are only usable on platforms that expose these architectural
...
+
+Code that depends on the contents of call-used registers surviving a function return may be incompatible, for example hand-written assembler, functions using a non-standard calling convention, or interfaces that return values in call-used registers outside of the platform ABI. Support is also target-dependent: Clang implements the option for AArch64, RISC-V, and x86 only[^clang-zero-call-used-regs].
```
thomasnyman: RISC-V support is in the upcoming Clang 23 version, GCC also supports [SPARC](https://gcc.gnu.org/pipermail/gcc-cvs/2020-December/338342.html)

```suggestion
Code that depends on the contents of call-used registers surviving a function return may be incompatible, for example hand-written assembler, functions using a non-standard calling convention, or interfaces that return values in call-used registers outside of the platform ABI. Support is also target-dependent: Clang implements the option for x86, AArch64, SPARC (GCC only), and RISC-V (since Clang 23)[^clang-zero-call-used-regs].
```

thomasnyman: Since the support is target-dependent, this option probably belongs in the situational options table under "for x86_64" and "for aarch64" in the tL;dr section.

thomasnyman (changes requested):
Thanks for the PR.

Overall this loks good although I'd like to see some clarifications to the individual option descriptions to improve readability. I also had some questions regarding the build failures with this option documented in Clang 17 (see below) and some clarification to the supported architectures.

I would also be very happy to see `used-arg`, `used-gpr-arg`, `all-arg`, and `all-gpr-arg` covered for completeness.

Regarding where in the tl;dr section this option should also be incorporated: I would agree it belongs in the situational options table due being supported only for specific target architectures. That can be a follow-up commit / PR.

thomasnyman:
@david-a-wheeler: Maybe you'd also have an interest in reviewing this contribution?

skypank-coder:
@thomasnyman Requested changes are addressed and about clang 17 , Yes — the two are different classes of bug. The Clang 15 issue was a code generation bug (NULL derefs at runtime, fixed in 16.0.0). The Clang 17 issue (llvm#69794) is an internal compiler error, i.e. the build fails loudly rather than producing bad code. It's confined to Clang 17: the reporter built the same OpenSSH release fine with Clang 16.0.5, upstream couldn't reproduce it on Linux, and OpenSSH's configure.ac excludes only 17*. I've clarified this in the section and footnote so the 16.0.0 recommendation doesn't read as contradictory.

thomasnyman (approved):
Thanks for implementing the requested changes, lgtm

thomasnyman:
I think this has had enough time for other to chime in with any concerns. Merging as is.

PR #1121, opened 2026-07-31
State-related code examples for pyscg-0052

s19110:
This change introduces new code examples that specifically show issues with the object's state not being properly reverted during an unexpected exception and show multiple solutions. It also updates the bibliography and the related guidelines section to use the HTML tables like other rules (also, the CWE pillar for this rule was missing).

For now, I have kept the lock example and removed the lock-related noncompliant and compliant code examples. I will investigate which base and pillar we can use for a rule that will include these and I will add such a rule in a separate PR.


docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0052/README.md
```diff
@@ -43,76 +43,223 @@ In the `noncompliant01.py` example, a thread gets locked, but not unlocked due t
 # SPDX-License-Identifier: MIT
 """Non-compliant Code Example"""
 
-import threading
+class DbConnection:
+    """Class representing a house"""
```
davidmather: This seems to be a copy-paste issue

docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0052/README.md
```diff
@@ -30,11 +30,11 @@ finally:
 
 ```
 
-It is best practice to use `with` statement in such cases as it will make sure the resource gets released even if an exception occurs in the execution.
+It is best practice to use `with` statement in such cases as it will make sure the resource gets released even if an exception occurs in the execution. There are other resources that also require cleanup, most commonly data streams for files, database connections, etc.
```
davidmather: The wording of the first few paragraphs needs work. Grok & ChatGPT have posted complaints about ambiguity of the wording or focus on the "with" statement when finally, is just as important

docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0052/README.md
```diff
@@ -30,11 +30,11 @@ finally:
...
 
 ## Non-Compliant Code Example
 
-In the `noncompliant01.py` example, a thread gets locked, but not unlocked due to an exception being thrown before it can be closed. This might lead to the lock remaining closed and inaccessible for further use.
+The `noncompliant01.py` contains a example of a stateful resource. The `DbConnection` class imitates a data stream, simplyfying it to the `connected` boolean value. It also provides the `connect` and `disconnect` methods for manipulating this value. The `read` method simulates an operation that results in an unexpected exception.
```
davidmather: Grammar issue should be "an example"

BartKaras1128:
```suggestion
The `noncompliant01.py` contains a example of a stateful resource. The `DbConnection` class imitates a data stream, simplifying it to the `connected` boolean value. It also provides the `connect` and `disconnect` methods for manipulating this value. The `read` method simulates an operation that results in an unexpected exception.
```

davidmather (commented):
General grammar issues found.
Grok is also suggesting instead of focusing only on "bad state", mention:

leaked file descriptors
deadlocks from unreleased locks
incomplete database transactions
sensitive temporary files
stale authentication sessions

These are real security and reliability consequences.

s19110:
Thank you both for the reviews!
@davidmather I have run the rule through ChatGPT and Gemini. Now, the wording should be gramatically correct, easier to follow, and emphasize `finally` more. Please let me know if anything else should be changed.

BartKaras1128 (approved):
Looks good to me now!

docs/Secure-Coding-Guide-for-Python/05_exception_handling/pyscg-0052/README.md
s19110: This is a leftover docstring for when I was exploring different code example ideas. It should be fixed now.

davidmather (approved):

Issue #806, opened 2025-02-27
pySCG: New Python specific rule that promotes signed software, CWE-494: Download of Code Without Integrity Check


myteron:
Supply chain attacks are the most efficient attack vector for malicious actors. We need a rule that outlines the issue, aka lack of software verification, and showcase how to at least verify already signed software.

[CWE-494: Download of Code Without Integrity Check](https://cwe.mitre.org/data/definitions/494.html)

Own signatures might require another rule to avoid overloading this one. 

Python modules are frequently hosted on mirrors with unknown trust.
Python.org provides a solution via Sigstore  since Python 3.10.7 
[Sigstore information](https://www.python.org/downloads/metadata/sigstore/)
it is not very well know amongst coders. 

Python allows to sign wheel files distribution format — Python Packaging User Guide

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#binary-distribution-format

ArockiaRajamanickam:
I'd like to write this rule, but the mechanism the issue points at has been deprecated since it was filed, so I want to agree the scope before drafting rather than after.

The issue links the binary distribution format spec for wheel signing. That spec now reads:

> `RECORD.jws` and `RECORD.p7s` are deprecated. [...] Build backends and other tools **must not** add them to wheels anymore

So a rule teaching wheel signatures would be teaching something tooling is forbidden to produce.

What replaced it is PEP 740 index attestations, which is Final and live on PyPI today. A package published through a trusted publisher carries a Sigstore bundle whose inclusion is recorded in Rekor, and PyPI exposes it at `/integrity/<project>/<version>/<file>/provenance`. That gives a verifier something the old scheme never did: the attestation says which repository and which workflow built the artifact, and the transparency log makes the claim non-repudiable.

Concretely, I'd propose the rule covers:

- **Noncompliant**: fetch a payload over HTTPS and `exec`/import/subprocess it, with no verification. HTTPS authenticates the *host*, not the *artifact*.
- **Noncompliant**: verify against a digest published on the same host as the artifact. Whoever replaces the file replaces the digest.
- **Compliant, stdlib only**: verify a digest obtained through a separately authenticated channel, using `hashlib` and `hmac.compare_digest` for the comparison, so it stays inside the Python Module Index scope the guide sets.
- **Prose**: what PEP 740 attestations give you, why the transparency log matters, and that TOFU pinning is a different trust model rather than a weaker version of the same one.

[...]

PR #1122, opened 2026-08-04
pySCG: add pyscg-0056 Verify Downloaded Code Before Running It

ArockiaRajamanickam:
Refs #806.

Adds `pyscg-0056: Verify Downloaded Code Before Running It` under Coding Standards, which is where you pointed me after the restructure.

## What it covers

The rule teaches one thing: an artifact must be checked against a value obtained through a channel the attacker does not control, before it is executed.

- **Noncompliant**: fetch a plugin over `https` and `exec` it. The exploit section substitutes different bytes to show that TLS authenticates the host that answered, not what it sent.
- **Compliant**: compare the artifact against a digest pinned in source control, using `hmac.compare_digest`. The substituted artifact is rejected before it reaches `exec`.

A short prose section then covers what a pinned digest does *not* give you, since that is the part a reader is most likely to get wrong: it proves the artifact has not changed since it was pinned, but says nothing about who produced it and does not scale to transitive dependencies.

## The mechanism in the issue is gone

The issue points at wheel signing. That mechanism has since been withdrawn — the binary distribution format specification now states that `RECORD.jws` and `RECORD.p7s` are deprecated and that build backends "must not add them to wheels anymore". Teaching it would have taught something tooling is forbidden to produce.

The rule describes PEP 740 index attestations instead, which is Final and live on PyPI today. I kept that to a few sentences rather than a full treatment, because it is index infrastructure rather than something the reader writes, and because you flagged in the issue that producing your own signatures belongs in a separate rule.

## On your two comments

**Keeping crypto examples short.** Both examples are under 30 lines and the compliant one uses only `hashlib` and `hmac`, so it stays inside the Python Module Index scope and needs no third-party install. There is no key handling, no certificate parsing and no signature format to explain — a digest comparison is the smallest thing that makes the point, which is why I did not reach for Sigstore verification in code.

**Placement.** In Coding Standards as you suggested. It reads as a coding habit rather than a cryptographic building block, so it did not seem to need the `09 Cryptography` grouping.

## Checks

`markdownlint-cli` with the repo's `.markdownlint.yml` is clean on both the new README and the updated `readme.md`. `ruff check --select E,F,W,S` and `ruff format --check` pass on both examples. Both scripts run: the noncompliant one executes the substituted payload, the compliant one rejects it. Every URL cited was fetched and returns 200 — including the SEI CERT cross-reference, where I had initially paired the right rule name with the wrong link. Rule ID 0056 is the next free number and does not collide with the IDs in the currently open PRs. The main `readme.md` table has a row for it, with CWE-494. Commit is signed off.

I used an AI assistant while working on this. The deprecation check against the packaging spec, the link verification and the decision to keep Sigstore out of the code examples are mine.

### DIFF
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: OpenSSF project contributors
+# SPDX-License-Identifier: MIT
+"""Compliant Code Example"""
+
+import hashlib
+import hmac
+
+# Obtained from the publisher through a channel the attacker does not control,
+# such as a pinned value in this repository. A digest served alongside the
+# artifact proves nothing: whoever replaces one replaces the other.
+EXPECTED_SHA256 = "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3"
+
+
+class IntegrityError(Exception):
+    """Raised when an artifact does not match its expected digest."""
+
+
+def verify(artifact: bytes, expected_hex: str) -> bytes:
+    """Return the artifact only if its digest matches the expected value."""
+    digest = hashlib.sha256(artifact).hexdigest()
+    if not hmac.compare_digest(digest, expected_hex):
+        raise IntegrityError("artifact does not match its expected digest")
+    return artifact
+
+
+#####################
+# Trying to exploit above code example
+#####################
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+try:
+    verify(ATTACKER_SOURCE, EXPECTED_SHA256)
+except IntegrityError as error:
+    print(f"rejected before execution: {error}")
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
@@ -5,10 +5,11 @@
 import hashlib
 import hmac
 
+PLUGIN_URL = "https://downloads.example.com/plugin.py"
 # Obtained from the publisher through a channel the attacker does not control,
 # such as a pinned value in this repository. A digest served alongside the
 # artifact proves nothing: whoever replaces one replaces the other.
-EXPECTED_SHA256 = "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3"
+EXPECTED_SHA256 = "9f6077fbddaa732bb91278098b727709ad4a3cf5c00022057b6775e77fb92127"
 
 
 class IntegrityError(Exception):
@@ -23,11 +24,32 @@ def verify(artifact: bytes, expected_hex: str) -> bytes:
     return artifact
 
 
+def get_plugin_source() -> bytes:
+    """Return the bytes served by PLUGIN_URL."""
+    return b"print('plugin loaded')\n"
+
+
+def load_plugin() -> dict:
+    """Download a plugin, verify it, and only then run it."""
+    namespace: dict = {}
+    try:
+        source = verify(get_plugin_source(), EXPECTED_SHA256)
+    except IntegrityError as error:
+        print(f"rejected before execution: {error}")
+        return namespace
+    exec(source, namespace)  # noqa: S102
+    return namespace
+
+
 #####################
 # Trying to exploit above code example
 #####################
-ATTACKER_SOURCE = b"import os\nos.system('id')\n"
-try:
-    verify(ATTACKER_SOURCE, EXPECTED_SHA256)
-except IntegrityError as error:
-    print(f"rejected before execution: {error}")
+# A mirror, a cache, or anyone able to alter the response body serves these
+# bytes instead. TLS authenticates the host that answered, not what it sent.
+def tampered_source() -> bytes:
+    """Stand in for a response body the attacker replaced."""
+    return b"import os\nos.system('id')\n"
+
+
+get_plugin_source = tampered_source  # noqa: F811
+load_plugin()
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
@@ -2,15 +2,17 @@
 # SPDX-License-Identifier: MIT
 """Non-compliant Code Example"""
 
-import urllib.request
-
 PLUGIN_URL = "https://downloads.example.com/plugin.py"
 
 
-def load_plugin(url: str) -> dict:
+def get_plugin_source() -> bytes:
+    """Return the bytes served by PLUGIN_URL."""
+    return b"print('plugin loaded')\n"
+
+
+def load_plugin() -> dict:
     """Download a plugin and run it."""
-    with urllib.request.urlopen(url) as response:  # noqa: S310
-        source = response.read()
+    source = get_plugin_source()
     namespace: dict = {}
     exec(source, namespace)  # noqa: S102
     return namespace
@@ -19,9 +21,12 @@ def load_plugin(url: str) -> dict:
 #####################
 # Trying to exploit above code example
 #####################
-# A mirror, a cache, or anyone able to alter the response body serves this
-# instead. TLS authenticates the host that answered, not the bytes it sent.
-ATTACKER_SOURCE = b"import os\nos.system('id')\n"
-namespace: dict = {}
-exec(ATTACKER_SOURCE, namespace)  # noqa: S102
-print("attacker code ran with the privileges of this process")
+# A mirror, a cache, or anyone able to alter the response body serves these
+# bytes instead. TLS authenticates the host that answered, not what it sent.
+def tampered_source() -> bytes:
+    """Stand in for a response body the attacker replaced."""
+    return b"import os\nos.system('id')\n"
+
+
+get_plugin_source = tampered_source  # noqa: F811
+load_plugin()
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
@@ -1,0 +1,27 @@
+# SPDX-FileCopyrightText: OpenSSF project contributors
+# SPDX-License-Identifier: MIT
+"""Non-compliant Code Example"""
+
+import urllib.request
+
+PLUGIN_URL = "https://downloads.example.com/plugin.py"
+
+
+def load_plugin(url: str) -> dict:
+    """Download a plugin and run it."""
+    with urllib.request.urlopen(url) as response:  # noqa: S310
+        source = response.read()
+    namespace: dict = {}
+    exec(source, namespace)  # noqa: S102
+    return namespace
+
+
+#####################
+# Trying to exploit above code example
+#####################
+# A mirror, a cache, or anyone able to alter the response body serves this
+# instead. TLS authenticates the host that answered, not the bytes it sent.
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+namespace: dict = {}
+exec(ATTACKER_SOURCE, namespace)  # noqa: S102
+print("attacker code ran with the privileges of this process")
--- a/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/README.md
+++ b/docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/README.md
@@ -1,0 +1,169 @@
+# pyscg-0056: Verify Downloaded Code Before Running It
+
+Verify a downloaded artifact against a digest or signature obtained through a separate channel before executing it.
+
+Transport security authenticates the host that answered the request, not the bytes it returned. A mirror, a caching proxy, a compromised build host, or anyone able to alter the response body can substitute code, and the receiving process runs it with its own privileges.
+
+Two mistakes recur:
+
+* Executing an artifact after fetching it over `https`, treating the transport as proof of provenance.
+* Comparing the artifact against a digest published beside it. Whoever replaces the artifact replaces the digest.
+
+The expected value must therefore come from somewhere the attacker does not control, such as a value pinned in source control or a signature that chains to a key established out of band.
+
+## Non-Compliant Code Example
+
+The `noncompliant01.py` code example downloads a plugin and executes it without verification.
+
+[*noncompliant01.py:*](noncompliant01.py)
+
+```py
+"""Non-compliant Code Example"""
+
+import urllib.request
+
+PLUGIN_URL = "https://downloads.example.com/plugin.py"
+
+
+def load_plugin(url: str) -> dict:
+    """Download a plugin and run it."""
+    with urllib.request.urlopen(url) as response:  # noqa: S310
+        source = response.read()
+    namespace: dict = {}
+    exec(source, namespace)  # noqa: S102
+    return namespace
+
+
+#####################
+# Trying to exploit above code example
+#####################
+# A mirror, a cache, or anyone able to alter the response body serves this
+# instead. TLS authenticates the host that answered, not the bytes it sent.
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+namespace: dict = {}
+exec(ATTACKER_SOURCE, namespace)  # noqa: S102
+print("attacker code ran with the privileges of this process")
+```
+
+The substituted code executes with the privileges of the running process.
+
+## Compliant Solution
+
+The `compliant01.py` code example compares the artifact against a digest pinned in source control and refuses to return it on mismatch. `hmac.compare_digest` performs the comparison in constant time.
+
+[*compliant01.py:*](compliant01.py)
+
+```py
+"""Compliant Code Example"""
+
+import hashlib
+import hmac
+
+# Obtained from the publisher through a channel the attacker does not control,
+# such as a pinned value in this repository. A digest served alongside the
+# artifact proves nothing: whoever replaces one replaces the other.
+EXPECTED_SHA256 = "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3"
+
+
+class IntegrityError(Exception):
+    """Raised when an artifact does not match its expected digest."""
+
+
+def verify(artifact: bytes, expected_hex: str) -> bytes:
+    """Return the artifact only if its digest matches the expected value."""
+    digest = hashlib.sha256(artifact).hexdigest()
+    if not hmac.compare_digest(digest, expected_hex):
+        raise IntegrityError("artifact does not match its expected digest")
+    return artifact
+
+
+#####################
+# Trying to exploit above code example
+#####################
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+try:
+    verify(ATTACKER_SOURCE, EXPECTED_SHA256)
+except IntegrityError as error:
+    print(f"rejected before execution: {error}")
+```
+
+The substituted artifact is rejected before it reaches `exec`.
+
+## Verifying Packages Installed From an Index
+
+A pinned digest establishes that an artifact has not changed since it was pinned. It says nothing about who produced it, and it does not scale to transitive dependencies.
+
+For packages installed from PyPI, [PEP 740] index attestations record which repository and which workflow built a distribution, signed through Sigstore and recorded in a public transparency log. PyPI serves them at `/integrity/<project>/<version>/<file>/provenance`. Because the log is append-only and publicly auditable, a publisher cannot later deny a signature it issued.
+
+Note that the older wheel signing mechanism is no longer available. `RECORD.jws` and `RECORD.p7s` are deprecated in the binary distribution format specification, which states that build backends "must not add them to wheels anymore" \[PyPA 2026\].
+
+Trust on first use, where the first key seen is pinned and later keys are compared against it, is a different trust model rather than a weaker form of the same one. It detects a key changing but cannot establish that the first key was ever the right one.
+
+## Automated Detection
+
+<table>
+    <tr>
+        <td>Tool</td>
+        <td>Version</td>
+        <td>Checker</td>
+        <td>Description</td>
+    </tr>
+    <tr>
+        <td>Bandit</td>
+        <td>1.8.3 on Python 3.13</td>
+        <td>B102</td>
+        <td>exec_used</td>
+    </tr>
+    <tr>
+        <td>Bandit</td>
+        <td>1.8.3 on Python 3.13</td>
+        <td>B310</td>
+        <td>urllib_urlopen</td>
+    </tr>
+    <tr>
+        <td>Ruff</td>
+        <td>0.12 on Python 3.13</td>
+        <td>S102, S310</td>
+        <td>flake8-bandit rules, detects the execution and the fetch but not the missing verification between them</td>
+    </tr>
+</table>
+
+## Related Guidelines
+
+<table>
+    <tr>
+        <td><a href="http://cwe.mitre.org/">MITRE CWE</a></td>
+        <td>Base: <a href="https://cwe.mitre.org/data/definitions/494.html">[CWE-494: Download of Code Without Integrity Check]</a></td>
+    </tr>
+    <tr>
+        <td><a href="http://cwe.mitre.org/">MITRE CWE</a></td>
+        <td>Class: <a href="https://cwe.mitre.org/data/definitions/345.html">[CWE-345: Insufficient Verification of Data Authenticity]</a></td>
+    </tr>
+    <tr>
+        <td><a href="https://wiki.sei.cmu.edu/confluence/display/java/SEI+CERT+Oracle+Coding+Standard+for+Java">[SEI CERT Oracle Coding Standard for Java]</a></td>
+        <td><a href="https://wiki.sei.cmu.edu/confluence/display/java/SEC06-J.+Do+not+rely+on+the+default+automatic+signature+verification+provided+by+URLClassLoader+and+java.util.jar">[SEC06-J. Do not rely on the default automatic signature verification provided by URLClassLoader and java.util.jar]</a></td>
+    </tr>
+</table>
+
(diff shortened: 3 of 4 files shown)


myteron (commented):
Hi @ArockiaRajamanickam
  
Thanks for picking this up — it's a solid start covering CWE-494.

I'd suggest to either:

1. Focus on HMAC as a cryptographic building block for integrity and authenticity verification. This would fit
well as a rule in section 09 (Cryptography), demonstrating proper keyed-hash message authentication including
the caveats around symmetric key management and scalability in distributed systems.
  
2. Explain how compliant attestation for `.whl` packages works per the [index-hosted attestations spec](https://packaging.python.org/en/latest/specifications/index-hosted-attestations/), demonstrating the Python ecosystem's current approach to signed software using ECDSA, SHA256 X.509 and OIDC/JWT.

Reason: We encountered a similar overlap before in [pyscg-0023: Secure Deserialization](https://github.com/ossf/wg-best-practices-os-developers/blob/main/docs/Secure-Coding-Guide-for-Python/04_neutralization/pyscg-0023/README.md). HMAC preserves integrity but relies on symmetric keys, which don't scale well for package distribution where you need to verify provenance from publishers you don't share secrets with. That's what digital signatures (PEP 740 / Sigstore) solve.
 
Rg myteron

ArockiaRajamanickam:
Thanks, that framing helps. I would go with option 2, the attestation route, and I think option 1 is a genuinely different rule.

The reason is scope: a rule about HMAC as a keyed-MAC building block stops being about CWE-494. It would be about message authentication between parties that already share a secret, which is worth writing but does not answer "should I run this thing I just downloaded". Option 2 stays on the original question and matches what you asked for in the issue, verifying already-signed software.

One clarification on the current draft, because it changes which of your concerns applies. `hmac.compare_digest` is not HMAC. It is documented as "Return `a == b`" and takes no key; it just compares in constant time. The keyed MAC is `hmac.new(key, msg, digestmod)`, which the draft never calls. So there is no symmetric key in the example and no key distribution problem to scale. What the draft does have is the weakness underneath your point: a pinned digest proves the artifact has not changed since somebody pinned it, and says nothing about who produced it. That is exactly why attestations are the better answer, so your conclusion holds even though the HMAC part does not apply.

Before I rewrite it, one constraint I would rather settle with you than guess at.

I pulled a live attestation to check what is actually achievable:

```
GET https://pypi.org/integrity/sigstore/4.5.0/sigstore-4.5.0-py3-none-any.whl/provenance  -> 200

publisher:              kind=GitHub, repository=sigstore/sigstore-python, workflow=release.yml
envelope:               statement, signature          (DSSE)
verification_material:  certificate, transparency_entries   (Fulcio X.509, Rekor)
predicateType:          https://docs.pypi.org/attestations/publish/v1
subject:                sigstore-4.5.0-py3-none-any.whl, digest sha256
```

Fetching that and reading the publisher identity is straightforward with `urllib`, `json` and `base64`. Actually verifying it is not: it needs the DSSE signature checked, the Fulcio certificate chained to a trusted root, and the Rekor inclusion proof validated. That is not reachable from the standard library, and every code example in the guide today is stdlib only.

So which do you prefer?

1. The compliant example fetches the attestation and reads the publisher identity with stdlib, and the rule states plainly that real verification is `pypi-attestations` or `sigstore` and shows the command rather than reimplementing it.
2. This rule takes a dependency on `sigstore` as a deliberate exception, since the subject matter cannot honestly be demonstrated without one.

I lean towards 1, because it keeps the guide's stdlib property and avoids showing beginners a hand-rolled verification they should never write. It also lets the rule make the point that the `subject.digest.sha256` in the attestation is what binds the signature to the artifact, so the digest comparison becomes a step inside verification rather than an alternative to it.

Happy to rewrite either way, and to move it to section 09 if you would rather it sat with the cryptography building blocks.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
```diff
@@ -0,0 +1,27 @@
...
+
+#####################
+# Trying to exploit above code example
+#####################
+# A mirror, a cache, or anyone able to alter the response body serves this
+# instead. TLS authenticates the host that answered, not the bytes it sent.
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+namespace: dict = {}
+exec(ATTACKER_SOURCE, namespace)  # noqa: S102
+print("attacker code ran with the privileges of this process")
```
s19110: The `PLUGIN_URL` constant and the `load_plugin()` functions are never used.

To imitate the attacker replacing the response body with malicious code, we can create a separate function for downloading the source, and then have it replaced with a lambda in the attack section. This way, the original source will technically be used, and the `load_plugin()` function will be used to perform the attack instead of the attacker manually running `exec`. I also think the print statement might be redundant since the console output will already contain the result of running the `ATTACKER_SOURCE` command. 
Here is my suggestion as to how it could be implemented:


```suggestion
"""Non-compliant Code Example"""

import urllib.request

PLUGIN_URL = "https://downloads.example.com/plugin.py"

def get_plugin_source() -> str:
    """Download the plugin source code"""
    with urllib.request.urlopen(PLUGIN_URL) as response:  # noqa: S310
        return response.read()

def load_plugin() -> dict:
    """Download a plugin and run it."""
    source = get_plugin_source()
    namespace: dict = {}
    exec(source, namespace)  # noqa: S102
    return namespace


#####################
# Trying to exploit above code example
#####################
# A mirror, a cache, or anyone able to alter the response body serves this
# instead. TLS authenticates the host that answered, not the bytes it sent.
get_plugin_source = lambda: b"import os\nos.system('id')\n"
load_plugin()

```

In this case, the `get_plugin_source` method serves as the host responding to the request, and the attacker is altering its response.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+#####################
+# Trying to exploit above code example
+#####################
+ATTACKER_SOURCE = b"import os\nos.system('id')\n"
+try:
+    verify(ATTACKER_SOURCE, EXPECTED_SHA256)
+except IntegrityError as error:
+    print(f"rejected before execution: {error}")
```
s19110: The exploit section of the code should usually remain unchanged from the noncompliant code example. This allows us to show that while the attacker tries the same attack method, it no longer works after applying suggested secure coding guidelines.
In the following suggestion, I have brought back the `load_plugin()` function and moved the verify check to its body. The suggestion also includes the changes I have suggested for the `noncompliant01.py` code example:


```suggestion
"""Compliant Code Example"""

import hashlib
import hmac
import urllib.request

# Obtained from the publisher through a channel the attacker does not control,
# such as a pinned value in this repository. A digest served alongside the
# artifact proves nothing: whoever replaces one replaces the other.
EXPECTED_SHA256 = "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3"
PLUGIN_URL = "https://downloads.example.com/plugin.py"

class IntegrityError(Exception):
    """Raised when an artifact does not match its expected digest."""


def verify(artifact: bytes, expected_hex: str) -> bytes:
    """Return the artifact only if its digest matches the expected value."""
    digest = hashlib.sha256(artifact).hexdigest()
    if not hmac.compare_digest(digest, expected_hex):
        raise IntegrityError("artifact does not match its expected digest")
    return artifact

def get_plugin_source() -> str:
    """Download the plugin source code"""
    with urllib.request.urlopen(PLUGIN_URL) as response:  # noqa: S310
        return response.read()

def load_plugin() -> dict:
    """Download a plugin and run it."""
    try:
        source = verify(get_plugin_source(), EXPECTED_SHA256)
        namespace: dict = {}
        exec(source, namespace)  # noqa: S102
        return namespace
    except IntegrityError as error:
        print(f"rejected before execution: {error}")
        return dict()


#####################
# Trying to exploit above code example
#####################
# A mirror, a cache, or anyone able to alter the response body serves this
# instead. TLS authenticates the host that answered, not the bytes it sent.
get_plugin_source = lambda: b"import os\nos.system('id')\n"
load_plugin()
```

Because this version of the code is quite lengthy, I also have an alternative suggestion in which we skip the `PLUGIN_URL` check. Since this is a mock URL, we never get a real response when running the code, so instead I hardcoded a random command that also is used to generate the `EXPECTED_SHA256`. This way, we can also show that if we comment out the first line of the attack section, we actually execute the legitimate code from the plugin:

```suggestion
"""Compliant Code Example"""

import hashlib
import hmac

# Obtained from the publisher through a channel the attacker does not control,
# such as a pinned value in this repository. A digest served alongside the
# artifact proves nothing: whoever replaces one replaces the other.
EXPECTED_SHA256 = hashlib.sha256(b"print('Hello World')").hexdigest()
PLUGIN_URL = "https://downloads.example.com/plugin.py"

class IntegrityError(Exception):
    """Raised when an artifact does not match its expected digest."""


def verify(artifact: bytes, expected_hex: str) -> bytes:
    """Return the artifact only if its digest matches the expected value."""
    digest = hashlib.sha256(artifact).hexdigest()
    if not hmac.compare_digest(digest, expected_hex):
        raise IntegrityError("artifact does not match its expected digest")
    return artifact

def get_plugin_source() -> bytes:
    """Download the plugin source code"""
    return b"print('Hello World')"

def load_plugin() -> dict:
    """Download a plugin and run it."""
    try:
        source = verify(get_plugin_source(), EXPECTED_SHA256)
        namespace: dict = {}
        exec(source, namespace)  # noqa: S102
        return namespace
    except IntegrityError as error:
        print(f"rejected before execution: {error}")
        return dict()


#####################
# Trying to exploit above code example
#####################
# A mirror, a cache, or anyone able to alter the response body serves this
# instead. TLS authenticates the host that answered, not the bytes it sent.
get_plugin_source = lambda: b"import os\nos.system('id')\n"
load_plugin()
```

s19110 (changes requested):
The rule contents are overall good and explain the importance of having an idependent source of truth very well.

I have added a few suggestions to the code examples. They mostly aim unify the attack sections between both code examples as well as integrate unused constants and functions.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
ArockiaRajamanickam: You are right, and this was the weakest part of the example: the vulnerable function was declared and then bypassed, so the attack never actually went through the code the rule is about. Taken as suggested. `get_plugin_source()` is now the thing the attacker replaces, `load_plugin()` runs the exploit, and the redundant print is gone.

One change from your snippet. `get_plugin_source = lambda: ...` trips ruff E731 (assigning a lambda), and rebinding the name trips F811 either way, so I used a named `def tampered_source()` and assigned that with a `# noqa: F811`. The redefinition is deliberate here, and the file already carries `# noqa: S102`, so an explicit marker seemed more in keeping than a lambda.

I also dropped `urllib` entirely rather than keeping the unused-import shape. Checking the rest of the guide, pyscg-0056 was the only rule making a network call in an example, so this now matches the others and, as you noted, the example actually runs.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
ArockiaRajamanickam: Agreed on the principle, and it is the better teaching shape: the attack section is now byte for byte identical in both files, so the only thing that differs is whether `load_plugin()` verifies first. `load_plugin()` is back, with the `verify()` call inside it.

I went with your second, shorter variant, since it is self-contained and runnable, with one change. Deriving `EXPECTED_SHA256` with `hashlib.sha256(b"...").hexdigest()` at runtime would contradict the sentence directly above it, that the digest has to come from a channel the attacker does not control. A reader could reasonably copy that pattern and compute the expected digest from the very bytes they are trying to check. So it is a pinned hex literal, which is what the comment claims it is.

While making that change I noticed a mistake of my own that your snippet inherited: the old `EXPECTED_SHA256` was 40 hex characters, which is a SHA-1 digest length, not SHA-256. It happened to be SHA-1 of `b"test"`. It never showed up because the comparison is meant to fail in the demo, but in a rule about verifying digests it should not have been there. It is now the real SHA-256 of the known good source, so commenting out the single attack line runs the legitimate plugin and the check passes, which is the behaviour you were after.

Thanks for the review, it made the examples noticeably better.

ArockiaRajamanickam:
Thanks @s19110, both suggestions are in, each answered inline. The examples are better for it.

Summary of what changed:

- both files now share an identical attack section, so the same attack visibly succeeds in one and is rejected in the other
- the attacker replaces the function that returns the response body, so `load_plugin()` is the code path that actually runs, rather than being declared and bypassed
- `compliant01.py` regains `load_plugin()` with `verify()` inside it
- no network access in either example, matching every other rule in the guide, so both run as-is; commenting out the one attack line runs the legitimate plugin instead

Two deviations from the snippets, both explained in the threads: a named function instead of a `lambda` (ruff E731), and `EXPECTED_SHA256` stays a pinned literal rather than being computed at runtime, since computing it would contradict the comment above it about out-of-band provenance.

That last one turned up a real defect of mine that the snippets inherited: `EXPECTED_SHA256` was 40 hex characters, a SHA-1 length rather than SHA-256. It is now the genuine SHA-256 of the known good source.

Verified: `ruff check --select E,F,W,S` and `ruff format` clean on both files, markdownlint clean against `.markdownlint.yml`, the README code blocks diffed byte for byte against the `.py` files, and all four paths run as documented (each example with and without the attack line).

@myteron this also supersedes the open question about `hmac.compare_digest`, the code around it has changed shape.

myteron:
Hi @ArockiaRajamanickam 

apologies for the late response. 

I have done some research regarding supply chain attack prevention and found some interesting details that would most likely blow your PR completely out of proportion. So high risk of overloading this so please let us know if you any sort of deadline on finishing this PR for a project.

**Note regarding sticking to the standard library:**
The reason for this is that we want to follow the same principal as secure coding guides in other languages and also keep scope in check.  Pypi has today several thousand modules in active use and we don't want to be the ones writing documentation for them.  There is also the issue of providing recommendations on specific modules. Exceptions can be made for modules that are already referenced in docs.python.org that are basically considered part of the wider ecosystem for CPython and fully community driven (not a commercial based solution). If the solution is not available for current 3.9 simply mention that the provided code example is only available in >= 3.10 or 3.11
The `sigstore` module is only mentioned in https://docs.python.org/3/whatsnew/3.14.html changelog. 

**Note on pyscg to CWE relations:**
We had tried to have a 1:1 relation between pyscg rules and CWE and have been advised against it by the MITRE team. So pyscg rules to provide solutions best practices to avoid multiple weaknesses such as CWE-494 and CWE-509. I leave the choice to you.

**Regarding content of this PR:**
I would love to see a rule that gives some hands on practical set of commands that can be used to assure attestation verification end to end but don't want to overload the scope or your PR either. At this moment it appears that CPython is
having a solution gap that will be tricky to fill without overloading this PR.

I believe it is beneficial to the reader to see something similar to the attestation example you had provided in your comment using `GET https://pypi.org/integrity/sigstore/4.5.0/sigstore-4.5.0-py3-none-any.whl/provenance` last week in the
  introduction of this rule with a bit of context.

----

To keep your scope within reason I would broadly leave it at that for your PR and leave major changes to later. So below are thoughts on future improvements outside of your PR.

**End to end assurance options:**

As it looks we always end with a gap between installing CPython via os based packages such as `apt install -y python3 python3-venv` and installing sigstore `python3 -m pip install -U pip sigstore` on Debian.
  
There have been discussions on discussion.python.org but I don't know of anything officially documented in docs.python.org that would close that gap today.
  
Some verbose details on what I mean by later changes and the suggested reduced scope for a 09 Cryptographic functions learning path (my brain-dump):
- [Suggested Building blocks for rules in 09 Cryptography](https://github.com/myteron/wg-best-practices-os-developers/blob/pyscg_806_sigsoft/docs/Secure-Coding-Guide-for-Python/09_cryptography/issue_711_building_blocks_draft.md)
- [DRAFT for a 09 Cryptography introduction](https://github.com/myteron/wg-best-practices-os-developers/tree/pyscg_806_sigsoft/docs/Secure-Coding-Guide-for-Python/09_cryptography)
- [Some stats on malware in pypi from osv.dev](https://github.com/myteron/wg-best-practices-os-developers/blob/pyscg_806_sigsoft/docs/Secure-Coding-Guide-for-Python/09_cryptography/pypi_supply_chain_chart.md)

ArockiaRajamanickam:
No deadline at all, this is not tied to any project of mine, so please take the time you need. And thank you for the detail, the scoping notes are more useful to me than a quick yes would have been.

Pushed in 8419265, keeping to the reduced scope you described.

**Standard library only.** Understood, and the reasoning about not writing documentation for thousands of PyPI modules makes sense. The code examples were already stdlib only, `hashlib` and `hmac`, and they stay that way. No `sigstore` dependency, and no verification step that needs one.

**The attestation example is now in the introduction**, as you suggested, using the provenance endpoint rather than any third party module. I checked the live response again before writing it up, and there was one thing worth getting right: the in-toto statement is not a top level field, it is base64 inside `envelope.statement`. Decoded, it is a `https://in-toto.io/Statement/v1` whose subject carries the filename and its SHA-256, with `predicateType` `https://docs.pypi.org/attestations/publish/v1`, and the bundle names the publisher as `sigstore/sigstore-python` via `release.yml`.

I wrote it up around the point the rule is actually making: the claim is not "this digest sits next to this file", it is "this file, with this digest, was built by this workflow in this repository". I also stated plainly that verifying the signature on that envelope needs tooling beyond the standard library and is out of scope here, so a reader is not left thinking `curl` alone verified anything.

**On CWE.** Useful to know MITRE advised against 1:1, I had assumed the opposite. The rule already cited two, CWE-494 as Base and CWE-345 as Class, so it was not 1:1 already, and I have added CWE-829, Inclusion of Functionality from Untrusted Control Sphere, which fits the "runs code it did not verify" framing well.

One flag on your suggestion rather than silently dropping it: **CWE-509 is Replicating Malicious Code (Virus or Worm)**, which is about self replicating malware rather than unverified downloads, so I do not think it is the one you meant. If you did want it I will add it, I just did not want to cite something that would read oddly to someone following the link.

**On the wider 09 Cryptography work**, I read your building blocks draft and the PyPI malware stats. Agreed that the install-CPython-then-install-tooling gap is real and cannot be closed inside this rule. Happy to leave it here and help with that separately if it would be useful once the learning path takes shape.

markdownlint clean, all links checked, and the README code blocks still diff byte for byte against the `.py` files.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
s19110: Thank you for the update. I think this is a good and simple solution. The only problem is that `PLUGIN_URL` is only mentioned in the docstring now. 

None of our current rules make network calls, but we might want to introduce such rules. Instead of removing `urllib`, we could mock the request response. Here is an example:

```

"""Non-compliant Code Example"""

import urllib.request
from io import BytesIO
from unittest.mock import patch

PLUGIN_URL = "https://downloads.example.com/plugin.py"


def get_plugin_source() -> bytes:
    """Return the bytes served by PLUGIN_URL."""
    with urllib.request.urlopen(PLUGIN_URL) as response:  # noqa: S310
        return response.read()


def load_plugin() -> dict:
    """Download a plugin and run it."""
    with patch("urllib.request.urlopen", return_value=BytesIO(b"print('plugin loaded')\n")):
        source = get_plugin_source()
    namespace: dict = {}
    exec(source, namespace)  # noqa: S102
    return namespace


#####################
# Trying to exploit above code example
#####################
# A mirror, a cache, or anyone able to alter the response body serves these
# bytes instead. TLS authenticates the host that answered, not what it sent.
def tampered_source() -> bytes:
    """Stand in for a response body the attacker replaced."""
    return b"import os\nos.system('id')\n"


get_plugin_source = tampered_source  # noqa: F811
load_plugin()
```

However, adding response mocking makes the code example more complicated and introduces concept that are not directly related to the rule. An alternative solution would be to remove `PLUGIN_URL`.

@myteron, I'd like to know your thoughts on this.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
s19110: That is a good point that hardcoding `hashlib.sha256(b"...").hexdigest()` could suggest to the reader that this is how they should obtain the SHA digest instead of deriving it from an independent channel. In this suggestion, it was supposed to make it clear that `EXPECTED_SHA256` matches the return value of `get_plugin_source()`, but I see how that could be misleading. It's also good that the code example now uses a real SHA-256. Now, if we omit the attack section, we can see that `load_plugin()` manages to execute the expected code. Thank you for improving these code examples.

s19110:
Thanks @ArockiaRajamanickam for addressing my comments.

I left responses to your comments. Apart from my suggestion about `PLUGIN_URL`, the code examples look good to me now.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/noncompliant01.py
ArockiaRajamanickam: Agreed that a constant living only in a docstring is unsatisfying, and you are right to put it to @myteron rather than either of us just picking. I ran both of your options plus a third so the decision has evidence behind it rather than taste.

**Your mock version works.** Attack path prints `uid=...`, and with the attack line commented out it prints `plugin loaded`. One nit if it is chosen: the `patch(...)` line is 92 characters and trips `E501` at the 88 limit, so it needs wrapping.

Two things about it worth weighing, neither fatal:

- The mock sits **inside `load_plugin()`**, which is the function the rule is teaching. A beginner reads `unittest.mock.patch` as part of what looks like production code, and mocking is unrelated to CWE-494.
- In the attack path the patch is dead weight. Once `get_plugin_source` is replaced, `tampered_source` never calls `urlopen`, so the mock does nothing on the run that matters.

**A third option**, if the goal is only to stop `PLUGIN_URL` being decorative, is to make it a real default argument, with no mocking and no imports:

```python
PLUGIN_URL = "https://downloads.example.com/plugin.py"


def get_plugin_source(url: str = PLUGIN_URL) -> bytes:
    """Return the bytes served by url."""
    return b"print('plugin loaded')\n"
```

Both paths behave correctly, `ruff --select E,F,W,S` is clean, and the file keeps zero imports. Its own wart, so it is on the table honestly: `url` is accepted and then not used in the body, which a sharp reader will notice.

So the three trade off as: mocking is the most realistic but imports test machinery into taught code; removing `PLUGIN_URL` is simplest but loses the "this arrived over the network" framing that CWE-494 is about; the default argument keeps the framing and stays import-free but has an unused parameter.

I have not changed anything yet, since the decision is @myteron's. Say which you want and I will push it, including the wrapped line if it is the mock.

Separately, thank you for the compliant01 note. Glad the real SHA-256 and the pinned literal read better, and yes, commenting out the single attack line now runs the legitimate plugin, which was the behaviour you were after.

docs/Secure-Coding-Guide-for-Python/08_coding_standards/pyscg-0056/compliant01.py
ArockiaRajamanickam: Thanks. Marking this one resolved since the compliant example is settled, so the only thing still open on the PR is the `PLUGIN_URL` question in the other thread, which is waiting on @myteron rather than on either of us.

myteron:
Hi @ArockiaRajamanickam before we keep on going back and forward here in comments I was wondering if you could join one of our 'work' meetings or at least our slack channel. There are two each week.
* Wednesday day from 3:00 PM to 4:00 PM UTC : [https://zoom-lfx.platform.linuxfoundation.org/meeting/95686621028?password=c2a8e490-9847-4105-8f9b-5b57a0feab26](https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fzoom-lfx.platform.linuxfoundation.org%2Fmeeting%2F95686621028%3Fpassword%3Dc2a8e490-9847-4105-8f9b-5b57a0feab26&data=05%7C02%7Chelge.wehder%40ericsson.com%7C9599e3e10da744bbd2b908dee29858e7%7C92e84cebfbfd47abbe52080c6b87953f%7C0%7C0%7C639197343059472488%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=v6ct1lqzyykSLE4sN2KZl9Mho8cytb%2BKAlBfTN28oiA%3D&reserved=0)
* Friday from 2:00 PM to 3:00 PM UTC 
 [https://zoom-lfx.platform.linuxfoundation.org/meeting/98011141383?password=518c1832-2f13-4741-9ac2-cbb03fb54263](https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fzoom-lfx.platform.linuxfoundation.org%2Fmeeting%2F98011141383%3Fpassword%3D518c1832-2f13-4741-9ac2-cbb03fb54263&data=05%7C02%7Chelge.wehder%40ericsson.com%7Ce989616183e549e2c1db08dee2983c94%7C92e84cebfbfd47abbe52080c6b87953f%7C0%7C0%7C639197342998134774%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=ZdB9PZSx8q089tbqYDzdjpTEaAcGLbed%2FV68XSgJoSs%3D&reserved=0)

Rg Helge

ArockiaRajamanickam:
Hi Helge, thank you for the invitation, I really appreciate it.

Unfortunately both of those slots clash with another important meeting I have at the same time each week, so I cannot make the Wednesday or Friday calls at the moment. I am definitely interested in joining future ones though, and I would be grateful if you could let me know when the schedule changes or if there is a session on a different day.

In the meantime I am happy to keep going here on the PR. Whichever way you decide on the `PLUGIN_URL` question, I will push it straight away.

Thanks again.

PR #1125, opened 2026-08-18
Update _LIBCPP_HARDENING_MODE_FAST version support

garazdawi:
A version is needed in order for the parser in `main.py` to include libc++ in the requires section.


thomasnyman (approved):
lgtm

PR #1129, opened 2026-08-21
Issue 1027 : Reduce 0003 content to be more concise

BartKaras1128:
Made a lot of changes to pyscg-0003. Deleted all the example code examples and replaced them with small bulletpoint sections. Kept just 2 noncompliant and compliant code examples which I found to be easy to understand. Kept the opening paragraph also.


myteron (approved):
Looks good to me

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
andrew-costello, line 3:
```suggestion
Avoid mixing bitwise shifts with arithmetic operations. Use clear mathematical expressions instead to maintain predictable behavior, readability, and compatibility.
```

Small readability suggestion, don't think the 2 instead's read great, here's an alternative suggestion: Avoid mixing bitwise shifts with arithmetic operations. Use clear mathematical expressions instead to maintain predictable behavior, readability, and compatibility.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -13,207 +13,20 @@ The use of bit-wise operations for arithmetic or flag registers can reduce reada
...
-```
+## When Bitwise Is Appropriate
 
-Code `example05.py` demonstrates how Python is changing required storage in the background while keeping the class of type `int`. While this does prevent overflow in Python it can have some unexpected issues interacting with other languages. A mistake in a loop can cause Python to just continue to eat up memory while finding a natural ending in other languages.
+Bitwise operations are fine when data is actually a bit collection: flags/masks, serialization, binary protocols, cryptographic primitives. Keep such code localized, document the assumed bit width, and prefer helper functions for packing/unpacking.
 
 ## Non-compliant Code Example (Left Shift)
 
-Multiplication by `4` can be archived by a `2x` left. The `noncompliant01.py` code demonstrates an attempt to calculate `8 * 4 + 10` in one line.
+Multiplication by `4` can be archived by a `2x` left shift. The `noncompliant01.py` code demonstrates an attempt to calculate `8 * 4 + 10` in one line.
```
andrew-costello:
```suggestion
archived
```

Typo, should be changed to achieved

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
```diff
@@ -260,7 +73,7 @@ foo >>= 2
 print(foo)
 ```
 
-The expectation of the `>>= 2` right shift operator is that it fills the leftmost bits of `0011 0010` with two zeros resulting in `0000 1100` or decimal twelve. Instead a potentially expected `-12` in `foo` we have internal processing truncating the values from `-12.5` to `-13`.
+This prints `-13` instead of the expected `-12`. Python's right shift on negative values is arithmetic (sign-extending) and truncates toward negative infinity, not toward zero as integer division would in many other languages.
```
andrew-costello, line 76: Standard Python integer division would give -13, just wondering should expected -12 be argued, i think we should focus more on how you cant rely on Right shift for standard math.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
andrew-costello, line 90: Should this not be an int with //, as it is the same as bitwise shifting, as right shift can only be an int, and never a float.

docs/Secure-Coding-Guide-for-Python/03_numbers/pyscg-0003/README.md
andrew-costello, line 29:
```suggestion
Multiplication by `4` can be achieved by a `2x` left shift. The `noncompliant01.py` code demonstrates an attempt to calculate `8 * 4 + 10` in one line.
```

Small typo error, archived should be changed to achieved.

andrew-costello (commented):
Few discussion points, and typo's and phrasing.

PR #1139, opened 2026-09-16
Finding fixing ai - split by heading 2

david-a-wheeler:
People liked the idea of guidance on finding and fixing vulnerabilities using AI, but found 1 large document overwhelming. It turns out that splitting by heading level 2 works out decently well.

Merging this will mean people can see what it looks like overall. We can make changes to the generating process later, but this basic process (using pandoc) seems to be a reasonable starting point. It auto-breaks up the larger markdown file into many smaller files *with* navigation.


david-a-wheeler:
No DCO. Let's try again.

PR #1140, opened 2026-09-16
Add "Finding and Fixing Vulnerabilities Using AI"

david-a-wheeler:
Add a new guide/course on finding and fixing software vulnerabilities using AI, plus the tooling to generate and maintain it:

- Finding.md, the source document (from Google docs "save as" markdown)
- split and template.html, to turn it into chunked, navigable HTML under sections/ via pandoc.
- clean-finding.awk, a POSIX awk script that split runs first to clean up Google Docs export artifacts (a bogus flat table of contents, stray heading-id attributes, hard tabs, uneven trailing whitespace, loose list-marker spacing, bare URLs) so Finding.md passes markdownlint. It's idempotent and rewrites Finding.md in place. It's called by split.
- google-apps-script.js, run from the Google Doc to linkify citations against the bibliography and warn about duplicate citation IDs. I include it so it's recorded somewhere.
- add-staleness-check.md, notes on optionally adding CI to catch a stale sections/ directory and/or integrate its generation.


david-a-wheeler:
I propose merging this. This doesn't mean that the WGs have accepted this guide (though I hope they will!), but it means that everyone can see what heading2 divisions looks like. The hope is that dividing this document by heading 2 will mean that we'll have lots of smaller files that people can immediately jump to and use, and not be overwhelmed by a longer document.

david-a-wheeler:
@gkunz - what do you think, okay to merge so people can see what it could look like?

david-a-wheeler:
All: I'm thinking about merging this in directly, not because it's approved by a WG, but so that I can show the expected formatting to both the AI/ML WG and the Best Practices WG. I plan to *NOT* add any links to these files from best.openssf.org, so you'd have to already know it exists before seeing it.

david-a-wheeler:
Ok, I hope people don't mind, but I want to show what it'd look like to the AI/ML WG in 30 minutes, so I'm going to merge this, NOT because both WGs approve, but so that people can see what the guide could look like once broken at header2 level.

david-a-wheeler:
Merging so I can demo this!

PR #1142, opened 2026-09-17
Fix missing toc section page

david-a-wheeler:


david-a-wheeler:
This fixes a broken link & adds important license info. I'm going to add this now, so reviewers can see it properly.

PR #1144, opened 2026-09-18
Finding fixing updates

david-a-wheeler:
Simplify generated URLs and other improvements


david-a-wheeler:
Again, merging so people can more easily see the results as a demo.

